PR #5, opened 2015-04-06
FIX typo for binder < 10 >

jangxyz:
I found a typo for `binder < 10 >`, while declaring `col_10`. It should be `type_10`.

Besides this, is there any way to declare a callback with any number of arguments?
I've been playing with this for a while, and all I could do at the moment is just add more binder struct for each of arguments I need.

### DIFF
--- a/sqlite_modern_cpp.h
+++ b/sqlite_modern_cpp.h
@@ -538,7 +538,7 @@ namespace sqlite {
 			dbb.get_col_from_db(7, col_8);
 			type_9 col_9;
 			dbb.get_col_from_db(8, col_9);
-			type_9 col_10;
+			type_10 col_10;
 			dbb.get_col_from_db(9, col_10);
 
 			l(std::move(col_1), std::move(col_2), std::move(col_3), std::move(col_4), std::move(col_5), std::move(col_6), std::move(col_7), std::move(col_8), std::move(col_9), std::move(col_10));


aminroosta:
thanks for the fix.
the simple idea is to use a `std::tuple<T...>` and expand it into a variadic function call.

The **C++14** standard has a solution for this problem : [Compile-time integer sequences N3658](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3658.html).

but unfortunately visual studio 2013 doesn't support it :(
only clang 3.3+ and gcc 4.8+ ( [c++14 compiler and library shootout](http://cpprocks.com/c1114-compiler-and-library-shootout/) ) .

the other option would be to use  http://stackoverflow.com/questions/687490/how-do-i-expand-a-tuple-into-variadic-template-functions-arguments or http://stackoverflow.com/questions/7858817/unpacking-a-tuple-to-call-a-matching-function-pointer stackoverflow posts as a workaround.  

i'll play with those if i get some free time :)

Killili:
Im just looking at the other fork by @KnairdA and he has some nice additions regarding that stuff.And he already did that.

aminroosta:
@KnairdA implementation is clever 
```c++
template<std::size_t Count>
class binder {
private:
	template <
		typename    Function,
		std::size_t Index
	>
	using nth_argument_type = typename utility::function_traits<
		Function
	>::template argument<Index>;

public:
	template<
		typename    Function,
		typename... Values
	>
	static typename std::enable_if<(sizeof...(Values) < Count), void>::type run(
		database_binder& db,
		Function&        function,
		Values&&...      values
	) {
		nth_argument_type<Function, sizeof...(Values)> value{};
		db.get_col_from_db(sizeof...(Values), value);

		run<Function>(db, function, values..., value);
	}

	template<
		typename    Function,
		typename... Values
	>
	static typename std::enable_if<(sizeof...(Values) == Count), void>::type run(
		database_binder&,
		Function&        function,
		Values&&...      values
	) {
		function(std::move(values)...);
	}
};
```
i think maybe `run<Function>(db, function, values..., value);` in the middle of code should be `run<Function>(db, function, std::move(values)..., value);`

KnairdA:
Yes, the call to `run` during the non-final recursion stage could probably be improved. While I do not think that I have fully comprehended C++'s perfect forwarding and value reference particulars, `std::forward` should be an even better fit in this instance?

```c++
run<Function>(db, function, std::forward<Values>(values)..., std::move(value));
```

Values that are already passed as universal references would then simply be _forwarded_ to the next stage until they are finally moved to the target function scope. At least this seems to fit the example on [cppreference.com](http://en.cppreference.com/w/cpp/utility/forward).

Of course it could also be the case that it doesn't matter either way and the compiler is smart enough to collapse the recursion on it's own. It would certainly be interesting to compare the assembler output for both versions.

unphased:
Wonder if someone could give an example of how to provide a variadic template function or extract a tuple of values. [This](http://stackoverflow.com/questions/7858817/unpacking-a-tuple-to-call-a-matching-function-pointer) shows how to get a function called with its args being the tuple contents but I would be content just to have the tuple itself, e.g. as the return value of the statement itself. For example `db << "select id, name, count from table;"` might return to me a `tuple<int, string, int>`. That would feel natural I think.

Killili:
Yes there is definitely some confusion here, we do not parse the query we extract the type from the variables you give us after the SQL command. And that will happen on every command separately so if you have different types but the same names we will never know because your SQL is only read by SQLite.

unphased:
OK so there is a (necessary..) fundamental disconnect with the SQL types and C++ types, that is to say, because the SQL query itself can be dynamic in nature, the data assembly into the provided C++ lambda can only reasonably be done now on a best-effort basis. 

For example say that I'm using a websocket to send my SQL query into my C++ application from a webpage, I suppose what I can do is send in an enumeration of what I expect the types to be (and the count of the results), and then in C++ I will construct an appropriate lambda with the appropriate types for its parameters, and then feed that into `sqlite_modern_cpp`, then work from there. 

And it seems very lenient so I can (for now) freely use a lambda taking like 10 arguments, and have most of them be strings, and I can mostly get everything done through that...

Killili:
I dont know what you want to do but this lib is definitely not a good start for a generic SQL selection solution. And there is no way of "constructing" a lambda at runtime. They have to exist at compiletime already. 

This lib is a convenience layer atop SQLites own API, and not a replacement or ORM type of lib. 

If you need a generic SQL select use the original SQLite API type analysis capabilitys. There is a special constructor for our database that takes an already open SQLite connection so just open one yourself and use the original API for the "dynamic" stuff and when you need the added convenience of typesafty and style construct a temporary sqlite::database object. 

This is the way i do it and it works great.

unphased:
Insightful, thank you for your suggestions. Since my requirements are quite lax for now I can get by with some of the hacks I have mentioned. But, you are absolutely right. Poking into SQLite API a little bit, as you describe, sounds like a very much more robust approach! Thanks.

PR #9, opened 2015-05-17
Added some data-types

jgaa:
Added boost::optional support to deal with NULL values. If a value is unset during an insert,  NULL will be inserted to the database. When querying data, boost::optinal can retain the NULL state from the database.
Added boost::uuids::uuid (stored as blob) to ease work with distributed data
Added std::time_t support(stored as 64 bit integer / utc).
Added method to get the latetest insert ID.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ using namespace std;
 		// executes the query and creates a 'user' table
 		db <<
 			"create table if not exists user ("
+			"	_id integer primary key autoincrement not null,"
 			"	age int,"
 			"	name text,"
 			"	weight real"
@@ -38,6 +39,8 @@ using namespace std;
 			<< 21
 			<< "jack"
 			<< 68.5;
+			
+		cout << "The new record got assigned id " << db.last_insert_rowid() << endl;
 
 		// slects from user table on a condition ( age > 18 ) and executes
 		// the lambda for each row returned .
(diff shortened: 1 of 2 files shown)


aminroosta:
cool !
it will great to have `boost` support.

just some minor changes before merging this.

1- rename `LastInsertId` to `last_insert_rowid` (so it will be consistent with the rest of library).
2-(could you please) update the readme with boost a section and an example ?
3-and add a one line code example for `last_insert_rowid` in the first section of readme.

jgaa:
I have committed the modifications you suggetsed.

aminroosta:
Thank you @jgaa :)

Killili:
Hi, nice addition. But it does not compile on VS2013. Apperently std::time_t is typedefed to "long long" wich in turn is the same as sqlite_int64 and that is already there. Leading to this error:
`c:\libs\sqlite_modern_cpp\src\sqlite_modern_cpp.h(188): error C2535: 'sqlite::database_binder &sqlite::database_binder::operator <<(sqlite_int64)' : member function already defined or declared`

aminroosta:
thanks @Killili . from now on i will not merge any PR before checking it .

@Killili could you please document the new boost types you have added minutes ago ! (tnx).

one more thing , it's been a while we haven't released a new version.
i think it's time for another release. 

what do you think ? 

thanks **again**.

PR #14, opened 2015-10-21
Allow database binder to throw in destructor.

edrosten:
See long commit message.


aminroosta:
@edrosten  Thank you very much. excellent contribution :+1:

PR #17, opened 2015-11-18
Add test for try-catch-block

maschs:
@edrosten 
Does that look sufficient?

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -69,7 +69,7 @@ class sqlite3_statment
 		sqlite3_stmt* _stmt;
 
 	public:
-		
+
 		sqlite3_stmt** operator&()
 		{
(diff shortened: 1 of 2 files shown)


maschs:
@edrosten 
Your idea was correct, using `std::uncaught_exception()` fixed the bug. The new test is using a temporary file as test database now.

Killili:
@maschs  Thx

@aminroosta 
why is Travis-CI trying to compile us as a Ruby/Rake project?

maschs:
You don't have `.travis.yml` defining what kind of project this is. So Travis is running a default set.

Killili:
would be interesting, and i would never turn down help ;)

does/could it run the tests too?

maschs:
Ofc, its fully configurable. You can tell it which compilers to use (multiple, too) and basically you can run anything you'd like.

PR #19, opened 2015-12-08
fix "statment" to "statement"

unphased:
curiously enough, sqlite3.c has 2 instances of this typo


aminroosta:
Thanks for your contribution :+1:

PR #28, opened 2016-01-16
Ignore build products

tokuchan:
This commit updates git to ignore build-generated files and directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,10 @@
+
+.deps
+
+config.log
+
+config.status
+
+Makefile
+
+tests/*
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ config.status
 Makefile
 
 tests/*
+!tests/*.cc


Killili:
Hi,
i'm ok with most of this but why would we add "tests\" to ignore? We want those files in git.

tokuchan:
Good point. What about the generated test files? Should I just list them explicitly?

Sean Spillane
tokuchan@gmail.com

へ　へ
の　の
　も
　へ

Killili:
Best thing would be to have the build products stored somewhere else and ignore that path then i guess. Or exclude .cc files ... 

Its been a while since i build the project in its own directory so i dont know what files are generated.

tokuchan:
I can add `!tests/*.cc` to un-ignore the source files. That would address the issue.

Sean Spillane
tokuchan@gmail.com

へ　へ
の　の
　も
　へ

aminroosta:
Looks good to me, `!tests/*.cc` should be sufficient.

tokuchan:
Cool.

Issue #29, opened 2016-01-22
Ability to get sqlite3 handle in sqlite::database


mathieugarcia:
Hi,

I think it would be beneficial to have an accessor the the sqlite3 "db_" member in sqlite::database. I need to call a specific function but I don't want to mess with the sqlite3 initialization myself.

Is that something you think is worth adding ?

Thank you !

aminroosta:
Hi Mathieu,
If your specific function is a general problem that needs to be addressed by the library, you can explain it here and we will try to add it to the library.

the `sqlite3 * _db;` filed is private, if you really need this member field just make it public.
but don't send a PR for that, because i don't want to expose internal slqite3 stuff to the library users.

Thanks.

Killili:
Funny, i was just thinking about that. There are some cases i would like to get that handle out of the lib myself but i like that the lib is handling all the connection disconnection stuff.

So i would propose a change that would be more in line with the C++ ideology in terms of resources and RAII and capsule the sqlite3*  in a std::shared_pointer with a custom destructor to close the DB properly when nobody is using it. This way we can expose the sqlite3* to the outside world (with an accessor of course) and still be sure it will be handled correctly.

aminroosta:
Hi Killili,


What are the specific use-case you have for the handle? If there is a general problem we should address it via library itself.


I dont think making things complicated to access a member field will help average users. Professional users will probably hack their way into the library and add their custom tweaks.


But if there is a general problem or use-case we can definitely implement it.


Thanks ;)


Sent from Outlook Mobile



From: Killili

Sent: Friday, January 22, 6:37 PM

Subject: Re: [sqlite_modern_cpp] Ability to get sqlite3 handle in sqlite::database (#29)

To: aminroosta/sqlite_modern_cpp

Cc: amin roosta



Funny, i was just thinking about that. There are some cases i would like to get that handle out of the lib myself but i like that the lib is handling all the connection disconnection stuff.


So i would propose a change that would be more in line with the C++ ideology in terms of resources and RAII and capsule the sqlite3* in a std::shared_pointer with a custom destructor to close the DB properly when nobody is using it. This way we can expose the sqlite3* to the outside world (with an accessor of course) and still be sure it will be handled correctly.

Killili:
[...]

PR #31, opened 2016-01-23
Changed the way arguments are chained #29

Killili:
Major Changes and General Overhaul of the everything.

-Refactored
exceptions into its own namespace
-removed rvalue references where they
are not needed
-made prepared statments possible
-made them
reusable
-removed some template magic

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -99,8 +99,7 @@ namespace sqlite {
 
 	template<std::size_t> class binder;
 
-	//template<typename T> database_binder&& operator <<(database_binder&& db,T const&& val);
-	//template<typename T> void get_col_from_db(database_binder& db, int inx, T& val);
+	typedef std::shared_ptr<sqlite3> connection_type;
 
 	template<typename Tuple, int Element = 0, bool Last = (std::tuple_size<Tuple>::value == Element)> struct tuple_iterate {
 		static void iterate(Tuple& t, database_binder& db) {
@@ -133,6 +132,9 @@ namespace sqlite {
 				exceptions::throw_sqlite_error(hresult);
 			}
 		}
+
+		void set_used(bool state) { execution_started = state; }
+		bool get_used() const { return execution_started; }
 
 	private:
 		std::shared_ptr<sqlite3> _db;
@@ -286,7 +288,7 @@ namespace sqlite {
 			return database_binder::chain_type(new database_binder(_db, std::string(sql)));
 		}
 
-		std::shared_ptr<sqlite3> get_sqlite3_connection() const { return _db; }
+		connection_type get_sqlite3_connection() const { return _db; }
 
 		sqlite3_int64 last_insert_rowid() const {
 			return sqlite3_last_insert_rowid(_db.get());
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -5,6 +5,7 @@
 #include <stdexcept>
 #include <ctime>
 #include <tuple>
+#include <memory>
 
 #ifdef _MODERN_SQLITE_BOOST_OPTIONAL_SUPPORT
 	#include <boost/optional.hpp>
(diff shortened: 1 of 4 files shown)


Killili:
This is somewhat related to #29 but it got a bit out of hand so i made a new pull for this.
Its pretty much what i thought would be possible if we changed every lref of database_binder to a unique pointer so stuff like this is possible now:
```c++
try {

		{ // scope somewhere else
		  // creates a database file 'dbfile.db' if it does not exists.
			database db("dbfile.db");

			db << "drop table if exists test"; // still works as intended cause the database_binder dies at the semicolon
			db <<
				"create table if not exists test ("
				"   id integer primary key autoincrement not null,"
				"   val int"
				");";

		} // destroy db

		database db("dbfile.db");

		std::cout << "auto&\n";
		auto& pps = db << "select ?"; // get a prepared parsed and ready statment

		int test = 4;
		pps << test; // set a bound var

		pps >> test; // execute and reset statment

		pps << 4; // bind a rvalue
		pps >> test; // and execute again

		auto& pps2 = db << "select 1,2,3,4,5"; // multiple extract test

		pps2 >> [](int a, int b, int c, int d, int e) {
			std::cout << "L " << a << b << c << d << e << "\n"; // still works as intended
		};

	} catch(exception& e) { // and here all statments die...
		cout << e.what() << endl;
	}
```

aminroosta:
Changes are good :+1: 
But build is failing.

Killili:
Ok it took my ~20h but now it looks good!
There is a lot of template magic in the lib so having more builds would be nice.

I can confirm VC2015 and gcc-5.2.1 and gcc-4.8.2(travis) can someone test clang?

Readme follows... ;)

aminroosta:
@Killili i can confirm that all 5 tests ran successfully on clang 7.0.2
Great job :+1:

hdr/sqlite_modern_cpp.h
```diff
@@ -16,515 +17,480 @@
...
-			call_back();
+			if(hresult != SQLITE_DONE) {
+				exceptions::throw_sqlite_error(hresult);
+			}
 		}
 
-		if (hresult != SQLITE_DONE) {
-			throw_sqlite_error(hresult);
-		}
+		void set_used(bool state) { execution_started = state; }
```
aminroosta: instead of two functions overload `.used() // getter` and `.used(bool) // setter` functions

hdr/sqlite_modern_cpp.h
```diff
@@ -16,515 +17,480 @@
...
-		_ownes_db(true) {
-		_connected = sqlite3_open16(db_name.data(), &_db) == SQLITE_OK;
-	}
+		database(std::shared_ptr<sqlite3> db):
+			_db(db) {}
+
+		//template<typename T>
+		//database_binder::chain_type operator<<(const T& sql) const {
+		//	return database_binder::chain_type(new database_binder(_db,std::string(sql)));
+		//}
```
aminroosta: // remove the comment if we are not using it :)

hdr/sqlite_modern_cpp.h
```diff
@@ -16,515 +17,480 @@
...
+
+		database_binder::chain_type operator<<(const char* sql) {
+			return database_binder::chain_type(new database_binder(_db, std::string(sql)));
+		}
+
+		connection_type get_sqlite3_connection() const { return _db; }
```
aminroosta: rename to  to `.connection()` . @Killili  does a setter makes sense ? (`.connection(shared_ptr<sqlite3*> db)`)

aminroosta:
@Killili i have added some comments on the code, these are all just suggestions,
you are free to follow or keep names the way they are ;)

aminroosta:
@Killili Please go one more iteration on readme file ;)
I recommend writing "sql queries" that work for examples.
Please add your database backup example for shared_connections section ( its a useful example).
Also please draft a new release after you've merged this,
I think there have been so many great changes.

Killili:
Thx.

Those examples are valid SQL and work, maybe they are a bit short. But i like having examples that dont go through the whole "create table... yada yada" and get right to the point.
The backup routine is a bit long for an example but i will try.
I'm undecided on the getter/setter conventions, both have pros/cons.

Killili:
As much as i hate the word "iteration" in the context of documentation its done ;)

As for the connection setter: I think that would add inconsistencies to prepared statements, as they have a copy of the connection after they are created. Consider:

```c++
database db("orig");
auto ps = db << "select ...";

database db2("new");
db.connection(db2.connection());
auto ps2 = db << "select ...";

ps++; // this will run on DB "orig"
ps2++; // this will run on DB "new"
```
I dont get what you mean with "draft" a release so i leave that "honor" to you ;)

aminroosta:
@Killili  I just wanted to thank you again :wink: 
Also thanks for database backup example, really useful example :trophy:

PR #32, opened 2016-01-24
Blob support with `std::vector<T>`

aminroosta:


aminroosta:
@Killili please review this PR :)

aminroosta:
@Killili I don't have access to windows, could you aslo please test on visual studio 2013 or 2015?

Killili:
@aminroosta Yes i can test that.

I have to admit that it makes it pretty easy to use, but i still dont like idea of just supporting std::vector. This might be because i don't use it anywhere where i have SQL and BLOBs in one project.
On the other hand i had no luck making/finding a better solution, so why not ;)

aminroosta:
Thanks, i will hold on merging this for 2 or 3 days. maybe something better comes to our minds :wink: 
If that wasn't the case, then we will use this for now :)

Killili:
Example works on VS2015. I go searching for my thinking cap ... we could have a real discussion about that through a more convenient channel than this chat? Like skype or TS ...

aminroosta:
Yeah sure, i will send you an email :)

Killili:
This is basically what i have tried so far, this would catch vector and list of any POD or trivial structure and even more advanced containers like queue. Problem is it catches way to much even std::string fits this criteria and surprisingly should even work.

And it has been a nightmare to get the templates we have to integrate this.

```c++
template<typename T, typename V = T::value_type, typename = std::enable_if_t<std::is_class<T>::value && std::is_trivially_copyable<V>::value >  >
	database_binder& operator<<(database_binder::chain_type& db, const T& c) {

		auto iter = c.begin();
		auto tmp = std::vector<int8_t>(sizeof(T::value_type)*c.size());
		while(iter != c.end()) {
			auto val = iter++;
			std::copy(reinterpret_cast<int8_t*>(&val), reinterpret_cast<int8_t*>(&val + sizeof(T::value_type)), std::back_inserter(tmp));
		}

		// write tmp to DB

		++db->_inx;
		return db;

	}
```

aminroosta:
I am going to merge this for now.
To be honest i didn't get much time to play with template template parameters.
But with a couple of hours i didn't get anywhere close to a working program.

This will work for now, and when we get chance to implement the more general approach there will be no backward compatibility issues.

Issue #34, opened 2016-03-05
Shouldn't statement->execute() set used to true?


mqudsi:
Hello,

Thanks for this library, I just migrated a project over from CppSqlite (which I'd been maintaining) over to this. 

I have a question on prepared statements - is there a reason manually calling `execute` on a statement (and _not_ `statement++`) does not set used to true? From the documentation and going by logic, I assumed `execute()` would execute the statement and set it to used. `reset` would unbind parameters and set `used` to false. But that's not the case since it remains "unused" even after explicitly executing and will run once more before destructing unless `used(false)` is manually called.

Thanks.

Killili:
Hi, and thanks.

As far as i remember, reset will reset the bindings and if you want to reuse them, as is the case with prepared selects you might not want to use op++.

mqudsi:
Thanks for the quick reply. I guess my question wasn't too clear.

	ErrorCode DBWrapper::InsertBlock(int blockID, uint32_t overlapOffset /*= -1*/)
	{
		auto statement2 = *_masterDB << "INSERT INTO 'Overlaps'\
						(BlockID, OverlapOffset) VALUES \
						(?, ?);";

		return SafeSql(__FUNCTION__, [&] {
				statement2->reset();
				statement2 << blockID;
				statement2 << (int)overlapOffset;
				statement2->execute();
              			statement2->used(true);
		});
	}

In the above code, if `statement2->used(true);` is not commented out, when `SafeSql` returns (which just executes the lambda in a `try..catch`) and `statement2`'s destructor is called, it will execute once again (and fail because the bindings are invalid).

The destructor for `database_binder` checks if `execution_started` is false in order to determine if it should be run once more. But calling `execute()` does not set `execution_started` to true in the above case.

aminroosta:
@mqudsi This looks like a bug to me.
@Killili  Will #35 fix this?

Killili:
Ok after thinking a bit about it (has been a few weeks since i wrote that).
And i had to read a lot of old Delphi code today so have mercy. 
I think the intentional separation between execute and reset is the thing that i was thinking about and this "used" issue is an oversight. 

So go ahead pull the fix.

aminroosta:
@Killili Cool!
Please take another look at diff in #35, If it looks good to you please merge it :-)
I also set `ps.used(false)` on `ps.reset()`, that should be fine. right?

mqudsi:
@Killili no worries, thank you kindly.

PR #35, opened 2016-03-05
prevent prepared- statements from executing again when they go out of…

aminroosta:
… scope

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -122,6 +122,7 @@ namespace sqlite {
 			sqlite3_reset(_stmt.get());
 			sqlite3_clear_bindings(_stmt.get());
 			_inx = 1;
+			used(false);
 		}
 
 		void execute() {
@@ -132,6 +133,7 @@ namespace sqlite {
 			if(hresult != SQLITE_DONE) {
 				exceptions::throw_sqlite_error(hresult);
 			}
+			used(true); /* prevent from executing again when goes out of scope */
 		}
 
 		void used(bool state) { execution_started = state; }


Killili:
I think i did that on purpose so until i remember why we should run with it.

aminroosta:
Sure, No problem :-)
Although what @mqudsi described in #34 looks like a bug.

Killili:
Nope its even documented behavior ;)

aminroosta:
@Killili 
You are right, Although its not that clear in the docs :) 
But i think it means that users should manually set `ps.used(true)`.
 
```c++
// If you dont need the returned values you can execute it like this
ps.execute(); // the statement will not be reset!
// To disable the execution of a statment when it goes out of scope and wasn't used
ps.used(true); // or false if you want it to execute even if it was used
```
Is there a good reason not to set `ps.used(true)` at the end of `ps.execute()` call?
I don't see why this can not be the default behavior.

mqudsi:
@Killili  I don't think it's documented that way.
What the documentation says:

        // If you dont need the returned values you can execute it like this
        ps.execute(); // the statment will not be reset!

"Will not be reset" doesn't (to me) directly imply "will still be executed", as "reset" is mentioned in the docs as:

        // after a successfull execution the statment needs to be reset to be execute again. This will reset the bound values too!
        ps.reset();

That aside, the docs say "If the statment was executed once it will not be executed again when it goes out of scope."  Calling `statement->execute()` would heavily imply (some would say, definitively so) that the statement was, well, executed.

aminroosta:
@mqudsi I agree, that's why i think it should be the default behavior.
I want @Killili  to take another look at the diff.
I am also setting `ps.used(false)` on `ps.reset()` call. which should be fine.  
but i want to make sure. :-)

Issue #38, opened 2016-03-13
seems uint64_t not supported?


douyw:
Hi,
Thank you for this very convenient library! 
I just downloaded the latest code. When uint64_t inserting, the following link error occurs: 
```
error: undefined reference to `std::unique_ptr<sqlite::database_binder, std::default_delete<sqlite::database_binder> >& sqlite::operator<< <unsigned long long>(std::unique_ptr<sqlite::database_binder, std::default_delete<sqlite::database_binder> >&, unsigned long long const&)'
```
Is it necessary to write a special template for uint64_t?

douyw:
Oh! I changed uint64_t into int64_t, it get compiled.

Killili:
Youre right there is a template missing.

Killili:
I honestly thought there was no uint64 support because sqlite uses int64 for everything. But there is. So Pull #39 will fix this.

Killili:
#39 Fixed it for VS2015 but travis (g++ 5) complains about it. And its a good thing so i had some time to think about it.

SQLite does not provide API functions to set and get unsigned values from a database. We can fake that by casting signed types to unsigned types (like i did in #39). 
But if anybody else would read data from that database he would have no clue to guess the signednes of integers. So i dont think we should add this!

PR #39, opened 2016-03-13
Added missing support for uint64_t

Killili:


Killili:
I will close this because it makes the database inconsistent see #38

Issue #15, opened 2015-11-16
multiple definition of sqlite::database_binder'


adnan-emd:
Adding "#include <sqlite_modern_cpp.h>" multiple .cpp files causes multiple definition errors.
This is on 64-bit Kubuntu 15.04 using Netbeans 8.02.  Compile errors are attached.

[compile_errors.txt](https://github.com/aminroosta/sqlite_modern_cpp/files/35729/compile_errors.txt)

edrosten:
Should be fixed now. Please reopen if the problem persists.

adnan-emd:
Thanks for the fix and thanks for the great library, Edward!

MordechaiMaman:
hi I saw that you fixed this issue ([commit](https://github.com/aminroosta/sqlite_modern_cpp/commit/1309a0d1a624bb49063003578f2e2f7886c1f602)) so I'm not sure when exactly that happen but I'm having the same issue again..
if you could take a look at the functions:
``` cpp
template<> database_binder::chain_type& operator <<(database_binder::chain_type& db, const std::string& txt) 
template<> void get_col_from_db(database_binder& db, int inx, std::u16string & w) 
template <typename BoostOptionalT> database_binder::chain_type& operator <<(database_binder::chain_type& db, const boost::optional<BoostOptionalT>& val) 
template <typename BoostOptionalT> void get_col_from_db(database_binder& db, int inx, boost::optional<BoostOptionalT>& o) 
void operator++(database_binder::chain_type& db, int)
```
at lines 467, 477, 498, 511 and 527
that would be great !

Killili:
Hi,
sorry but a bit more information is needed here. 
What is actually going wrong? On which compiler? Under what circumstances? Are you using boost optional?...

MordechaiMaman:
right sorry... 
the same issue of redefine for some symbols caused by two include of the main header from two different obj files.
I'm using vc 2015 in the default settings when I got all the header paths and library right (these and boost)
and it doesn't matter if I'm using boost optional or not this is still happening...
```
[...]

PR #40, opened 2016-03-16
add a few inlines to solve redefine symbole errors

MordechaiMaman:


Killili:
This should fix #15 (just for reference)

Killili:
Thx

PR #42, opened 2016-03-26
support functors

### LINKED ISSUE #11: How to use a functor to replace the lambda one?
Hello,
Thank you for this great sqlite wrapper! It's very convenient.
While using the wrapper is delightful, a problem about eliminating similar code occurs.

The functor:
```c++
struct ItemResultHandler {
    explicit ItemResultHandler(QVariantMap& map_) : map(map_) {}
    void operator() ( int id, std::string name, double price, int qty ) {
         map.insert("id", id);
         map.insert("name", QString::fromStdString(name));   
         map.insert("price", price);
         map.insert("qty", qty);
    }
    QVariantMap& map;
};
```
The code:
```c++
QVariantMap getItemById(int id) {
  QVariantMap map;
  database db(dbfile);
  db << "select id,name,price,qty from items where id=?"
     << id 
     >> ItemResultHandler(map);
  return map;
}
```
It has compiling errors:
```
../../myapp/src/utility/function_traits.h:12: error: type 'void (ItemResultHandler::*)(std::__1::basic_string<char>, std::__1::basic_string<char>, int, int)' cannot be used prior to '::' because it has no members
        decltype(&Function::operator())
                  ^
```
```
 ../../myapp/src/sqlite_modern_cpp.h:218: error: incomplete definition of type 'sqlite::utility::function_traits<ItemResultHandler>'
                        binder<traits::arity>::run(*this, func);
                               ~~~~~~^~
```
I have posted a question in SO. Could you please have a look at it? Thank you in advance.
http://stackoverflow.com/questions/30387896/how-to-eliminate-multiple-similar-sql-queries-with-one-resultset-hander

aminroosta:
Trying to fix #11 :-)

### DIFF
--- a/hdr/sqlite_modern_cpp/utility/function_traits.h
+++ b/hdr/sqlite_modern_cpp/utility/function_traits.h
@@ -31,5 +31,26 @@ namespace sqlite {
 			static const std::size_t arity = sizeof...(Arguments);
 		};
 
+    /* support the non-const operator ()
+     * this will work with user defined functors */
+		template <
+			typename    ClassType,
+			typename    ReturnType,
+			typename... Arguments
+		>
+		struct function_traits<
+			ReturnType(ClassType::*)(Arguments...)
+		> {
+			typedef ReturnType result_type;
+
+			template <std::size_t Index>
+			using argument = typename std::tuple_element<
+				Index,
+				std::tuple<Arguments...>
+			>::type;
+
+			static const std::size_t arity = sizeof...(Arguments);
+		};
+
 	}
 }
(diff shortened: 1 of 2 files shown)


Killili:
The new test compiles and runs in VC2015.
I tried to get an understanding of the fault but i'm lost in the template jungle... i have the suspicion that there is a easier way but i might be totally wrong here. Shouldn't std::function in itself take care of the lambda/function/functor/member stuff?

But on the other hand this works, so you could pull this.

aminroosta:
@Killili  Thank you for testing on VS :+1: 
I was looking at `function_traits` the other day, and i saw that it specializes on `operator() const` member function, but user defined functors usually just have `operator()` without the `const` keywoard.  

I tried specializing it without the `const` keyword and compiler seems to be picking the write specialization in case of both lambdas and functions.

PR #44, opened 2016-04-07
update sqlite version to 3.12.0

aminroosta:


aminroosta:
@Killili I read the [sqlite release history](http://www.sqlite.org/changes.html) and there are lots of improvements from version **3.8.5** to **3.12.0**.
I couldn't find breaking changes that could affect the library, if you think the same please merge this PR :-)

PR #54, opened 2016-06-19
trying to fix exceptions

### LINKED ISSUE #53: Cannot catch constraint exception
I'm trying to insert a row that produces a primary key constraint failure but the program crashes before I can catch the exception.

try {
        db << "INSERT INTO tb_mytable (mytable_id,description) VALUES (?,?)" << 1 << "cannot catch exception";
// inserting again to produce error
        db << "INSERT INTO tb_mytable (mytable_id,description) VALUES (?,?)" << 1 << "cannot catch exception";
    } catch (exception e)
    {
        cerr << e.what() << endl;
    } catch (...)
    {
        cerr << "this does not work either" << endl;
    }

libc++abi.dylib: terminating with uncaught exception of type sqlite::exceptions::constraint: constraint failed
The program has unexpectedly finished.

aminroosta:
I've added a new test for exceptions and tried to fix #53

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,16 +99,16 @@ It is possible to retain and reuse statments this will keep the query plan and i
 		ps >> [&](int a,int b){ ... };
 
 		// after a successfull execution the statment needs to be reset to be execute again. This will reset the bound values too!
-		ps->reset();
+		ps.reset();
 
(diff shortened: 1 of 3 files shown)


Killili:
Looks like a solution/workaround that can be used till we get something better.

Issue #58, opened 2016-08-29
Archlinux integration !


funilrys:
Hello,
Just for information I maintain the package of SQL Modern CPP in AUR !
You can find it [here](https://aur.archlinux.org/packages/sqlite_modern_cpp/) or for arch user `yaourt -S sqlite_modern_cpp`

Hava nice day !

aminroosta:
Cool :wink:

I've added a package managers section to the readme.
I'm optimistic that other users will probably create packages for other package managers like nuget or conan as well.

@funilrys Thanks for your contribution 👍

aminroosta:
@funilrys We have new major release :-)

funilrys:
@aminroosta I'm working on the update :)

zauguin:
@funilrys Please take a look at #90, the current release might be broken on some systems.

funilrys:
@zauguin that's why I test all on local before pushing to the world ;)

funilrys:
@aminroosta @zauguin AUR version updated to v3.1

zauguin:
Awesome!

aminroosta:
@funilrys We have new release [v3.2](https://github.com/aminroosta/sqlite_modern_cpp/releases/tag/v3.2), a feature-full and polished release ;-)

funilrys:
@aminroosta Thanks I'm working on it :smiley_cat:

funilrys:
@aminroosta AUR version updated to v3.2

PR #59, opened 2016-08-29
Add package managers section to readme

aminroosta:
Thanks to @funilrys.
Related Issue #58

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ int main() {
 ```
 
 Prepared Statements
-=====
+----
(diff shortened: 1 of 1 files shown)


funilrys:
Your welcome @aminroosta  ! have you think about creating a "Release notification" isssue ? It'll be much easier for those who maintain packages :) I saw it [here](https://github.com/linuxdeepin/developer-center/issues/41)  and it's really a good idea cause we just have to suscribe to issue ! 

It's just a suggestion so hava good night !

Issue #71, opened 2017-01-19
open as readonly


alext234:
I am interested to use 1 sqlite database file shared between 2 programs, 1 as a writer and 1 as a reader. Sometimes I have the exception `locked` or `busy` thrown at the writer program.  
Do you plan to support with flags as those from `open_v2` here?
https://www.sqlite.org/c3ref/open.html

aminroosta:
@alext234 please try #72, i think this should enable you to pass the parameters you want.

I'm not planning on merging this to master yet, not without a unit-test and some documentation.
If you are interested please post a sample code of which parameters you used for `flags` option.
Enough to give me an idea on how to write a unit test :-)

mtissington:
I too would like to see the ability to pass the flags in and open the database as read only.

aminroosta:
I think we need a uniform design for this & #77

zauguin:
@aminroosta I tried to implement such a unified design with #77 in #93.
Tests and documentation are still missing.

PR #72, opened 2017-01-19
support flags option in open_v2()

aminroosta:
#71


zauguin:
@aminroosta I think we should decide if we keep supporting UTF-16 names and convert them or if we require UTF-8 database names like SQLite if you pass flags. If we require UTF-8 this would apply for all additional options like encryption too, but especially for encryption UTF-8 is required for the password anyway.

zauguin:
Closing this because the functionality is provided dy #93.

PR #75, opened 2017-01-31
std::optional support

zauguin:
This is almost like boost::optional, but part of C++17.
So we have the boost::optional interface without a boost dependency.

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -8,6 +8,10 @@
 #include <memory>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


aminroosta:
@zauguin Ah, wonderful! Thanks for the PR :-)

I've added you as a collaborator to the project ([invite link](https://github.com/aminroosta/sqlite_modern_cpp/invitations)).
If you had time please update the README will an example :-)

zauguin:
Thanks. I will wait with the README change until the first compilers actually ship <optional>, until then it's more of a theoretical feature.

PR #76, opened 2017-02-01
std::vector allocator support

zauguin:
Right now we are arbitrary limited to the standard allocator. Let's fix this.

This is useful in combination with libraries like Botan or Intel TBB, which implement custom allocators.


aminroosta:
Awesome. Thanks @zauguin

Issue #10, opened 2015-05-19
Development and Releases


Killili:
Hi,
on the subject that came up in the last merge comments so everybody sees this.

I too think we should release a new version because so much has changed since last time.
Any suggestions what we should do for the next release or what we can do better?

aminroosta:
It's been a while i wanted to integrate [SQLCipher](https://github.com/sqlcipher/sqlcipher).
Also supporting other databases (mysql for example) will be another option to consider.

what do you think ?

Killili:
The SQLCipher stuff looks interesting but i dont think i will need it.

As for more DBs i dread the incompatibilitys but i welcome more options, and indeed i will need mysql or postgress support in a few month ... 

Whats youre opinion on breaking changes? I'm a fan of major breaks every major release so we can keep it modern. So this version would be v1.5 but if we had introduced the new operators and removed the current operator>> it would have bin v2.

aminroosta:
That does make sense, major versions would be because of major API changes.

and yes _incompatibilitys_ will be a big challenge .

polesapart:
sqlcipher just needs an extra call (to set the encryption keyword), everything else is compatible to sqlite.

Killili:
So it took almost a year but now we got a new Version. ;)

toonetown:
Seems like there have been quite a few changes since the 3.2 release - time for another one? 😉

aminroosta:
```c++
// overload >> for Rvalue containers (1)
auto numbers = db << "select numbers from tbl where id = ?" << 1 >> vector<int>();
// support operator >> for Rvalue containers of std::tuple (2)
auto poeple = db << "select age,name from tbl;" >> vector<tupe<int,string>>();

// single row blob (3)
vector<int> vblob;
db << "select numbers from tbl where id = 1" >> blob_t(vblob);
// multiple row int
vector<int> vrows;
db << "select age from tbl;" >> vrows;
```

For `v4` i was thinking to get 1,2 & 3 implemented, cause 3 is a breaking change.
(The interface above is "my" summery of #122)

If anyone wants to work on the above please give me a heads up, so we don't work on the same thing :-)
Let's put end of October as the deadline! If we have the above i'll release it as `v4` otherwise it'll be `v3.3`.

aminroosta:
The deadline is passed, i still want to get those features in before a new release but lack the motivation to work on it.

[...]

PR #77, opened 2017-02-01
[RFC] Minimal SQLCipher support

zauguin:
Is there any interest in SQLCipher support?
How should this be tested? I don't really want to introduce a SQLCipher dependency just for the tests.


aminroosta:
@zauguin I have no experiment with SQLCipher.

Here is @polesapart comment about this issue (from #10)

`
IIRC sqlcipher just needs an extra call (to set the encryption keyword), everything else is api   
compatible to sqlite, i.e. you can replace sqlite3.c & sqlite3.h from those from sqlcipher and insert   
the extra call in the right sequence and nothing else is required for compiling. But for link you need   
the openssl's libcrypto.
`

hdr/sqlite_modern_cpp.h
```diff
@@ -340,6 +340,34 @@ namespace sqlite {
...
+		database(const std::string &db_name, const std::string &key): database(db_name) {
+			set_key(key);
+		}
+		database(const std::u16string &db_name, const std::string &key): database(db_name) {
+			set_key(key);
+		}
+
+		void set_key(const std::string &key) {
+			if(auto ret = sqlite3_key(_db.get(), key.data(), key.size()))
+				exceptions::throw_sqlite_error(ret);
```
aminroosta: Please add a custom exception class for sqlcipher.

hdr/sqlite_modern_cpp.h
```diff
@@ -340,6 +340,34 @@ namespace sqlite {
 			return sqlite3_last_insert_rowid(_db.get());
 		}
 
+#ifdef SQLITE_HAS_CODEC
+		database(const std::string &db_name, const std::string &key): database(db_name) {
+			set_key(key);
+		}
+		database(const std::u16string &db_name, const std::string &key): database(db_name) {
```
aminroosta: @zauguin Make the second parameter of the constructor an instance of `struct sqlite_config { std::string key; };`. The reason i suggest this is that we might pass other flags in the future, an example is #72 which i've implemented in #71 but is not merged yet.

hdr/sqlite_modern_cpp.h
zauguin: There is no such thing.
SQLCipher isn't very good at error reporting.
For example if the key is wrong, it still returns SQLITE_OK and fails only if you try to access the database for the first time with the existing SQLITE_NOTADB.
The key and rekey functions only fail, if you pass them nullpointers, but this can't happen here.
So actually this are noops, but the functions are a part of sqlites more generic codec interface and other implementations may choose to fail here, so we should at least check the return code.

hdr/sqlite_modern_cpp.h
```diff
@@ -318,6 +323,18 @@ namespace sqlite {
...
 			_db(db) {}
 
+		database(const std::string &db_name, const sqlite_config &config):
+			database(std::u16string(db_name.begin(), db_name.end()), config) {}
+
+		database(const std::u16string &db_name, const sqlite_config &config): database(db_name) {
+#ifdef SQLITE_HAS_CODEC
+			if(!config.key.empty()) set_key(config.key);
+#else
+			assert(config.key.empty() && "Encryption supported is disabled.");
```
zauguin: We could use an exception instead, but passing a key without encryption support is IMO a strong indication, that the application is compiled with the wrong settings.

polesapart:
I use sqlcipher - by including it's header instead of sqlite and arranging proper ldflags - and using PRAGMAS instead of it's extended C api. While I don't see any problems with conditionally adding that support to sqlite_modern_cpp, it's not strictly necessary either.

Please note that sqlcipher is not the only project using SQLITE_HAS_CODEC and relying on it for autodetection may lead to incompatibilities in case someone uses one of these other extensions to sqlite.

zauguin:
SQLITE_HAS_CODEC is used for autodetection, because I think this works for every system using SQLITE_HAS_CODEC.
But I can't test it with other projects using this feature right now.

polesapart:
At least "see" (sqlite's own encrypted version, which is proprietary and requires license) uses SQLITE_HAS_CODEC for the same purposes. One wouldn't be able to use "see" transparently with this library, as it has no API compatibility with sqlcipher. I suggest you disable autodetection and require the user to pass a specific define, i.e. HAVE_SQLCIPHER or something.

aminroosta:
@polesapart @zauguin We are adding new `#define` for optional features.
We already have `_MODERN_SQLITE_BOOST_OPTIONAL_SUPPORT
` and now `ENABLE_SQLCIPHER` (which should be renamed to `_MODERN_SQLITE_ENABLE_SQLCIPHER`).

I wonder if you think there is an easy way to include new `.h` files that extend the core library.
Adding a `hdr/sqlite_modern_cpp/sqlcipher.h` should enable sqlcipher and a similar header to enable boost support.

aminroosta:
@zauguin How about `struct sqlcipher_database : public sqlite:database { ... }` in the new `hder/sqlite_modern_cpp/sqlcipher.h` file? Do you think that makes sense ?

zauguin:
I like this approach, but if we want to add additional features like this we may need virtual inheritance so that the features can be combined.

aminroosta:
@zauguin Nicely done! Looks good to me 👍 
@polesapart Any comment before we merge this?

aminroosta:
@zauguin Is it possible (& easy) to write a test for this?

zauguin:
Writing a test is easy but running it isn't. We could either require SQLCipher for our test suite or only test it if SQLCipher is available. I prefer the second option, but I don't know how to implement it with autotools. We might have to change our testing system too, because the other test should not run against SQLCipher.

aminroosta:
@zauguin I see that you have already implemented [Skipped tests to Makefile.in](https://github.com/aminroosta/sqlite_modern_cpp/pull/86/commits/9e37e6de118b2902b296b500f3678e813d8f6d93).

Add a test for sqlcipher and skip it by default in Makefile.in.

zauguin:
I added the test which only executes if ENABLE_SQLCIPHER_TESTS is defined and added a hack for travis to check using SQLite3 and SQLCipher.

polesapart:
@aminroosta  I came late to the party, but everything seems ok from where I stand. Thanks to @zauguin for his efforts!

PR #83, opened 2017-02-09
Fix UTF-8 encoding of filenames and SQL statements

### LINKED ISSUE #82: UTF-16 vs UTF-8 encodings
While convenient for windows users, UTF-16 is, generally speaking, less convenient from a portability point of view. On Linux UTF-8 is more common. Not sure about MacOS.

Currently, sqlite_modern_cpp prefers to use UTF-16 everywhere, from filenames to database encoding and stored strings themselves, and tries to convert from "standard" encoding to u16string using it's iterator construtor.

There are a few shortcomings from these decisions, some of which are manageable and some I consider to be bugs. Let's see:

1. From sqlite wording, regarding database creation: "The default encoding will be UTF-8 for databases created using sqlite3_open() or sqlite3_open_v2(). The default encoding for databases created using sqlite3_open16() will be UTF-16 in the native byte order." And from pragma encoding docs: "Once an encoding has been set for a database, it cannot be changed". The main concern here is that even windows users might prefer the database to store data in utf-8 as most strings are single-byte representable and storing them in UTF-16 takes more disk space. The ideal scenario would be for the user to select the database encoding at database creation time. I do believe (didn't test) that there's still time for user to run db << "PRAGMA encoding = \"UTF-8\""; right after opening the database for the first time, as sqlite uses deferred creation - but this should be documented anyway.

2. In several instances, from accepting the database name in constructor to sending strings to sqlite api, there's a std::string -> std::u16string conversion using u16string's constructor that accepts a char-by-char  iterator, as in std::u16string(orig.begin(), orig.end()). The problem is that std::string charset is undefined, and unsafe for anything other than ASC-II, so while std::string could be e.g. utf8 (as in Linux, depending on user locale), it maybe anything on windows, from CP_ACP to CP_OEMCP), and storing one and retrieving other may lead to hard to track problems.
I'm not sure which would be best here: to have explicit std::u8string overloads which pass utf8 direct to sqlite (which happily expects utf-8), or fix the conversions to utf16 (I don't think using std::u8string and iterators is enough for converting, from what I've read, something like std::wstring_convert or std::codecvt_utf8 might be required). I'm also not sure what to assume with standard std::string .

While the code simply works for most scenarios, specially when the language is english and no special chars are involved, these are serious concerns for i18n-aware applications. Also, my programs all uses utf-8 with multibyte chars and it seems to work - but I didn't conduce enough tests to see if sqlite is handling the queries correctly, i.e. if it's matching the text as it should). Perhaps I'm wrong and std::u16string(u8str.begin(), u8str.end()) actually converts the encoding - but I didn't find any docs suggesting so.

Sorry to put so many points in a single report, but I think these can be further distilled on the discussion.

zauguin:
Fixes #82.
This patch breaks compatibility for non-ASCII database names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@ sqlite modern cpp wrapper
 
 This library is a lightweight modern wrapper around sqlite C api .
 
+**Heads up!**
+The current master contains a _breaking change_ compared to the 2.4 release for database names with non-ASCII characters. For details, see [issue 82](https://github.com/aminroosta/sqlite_modern_cpp/issues/82).
+
 ```c++
 #include<iostream>
 #include <sqlite_modern_cpp.h>
--- a/README.md
+++ b/README.md
@@ -5,9 +5,6 @@ sqlite modern cpp wrapper
 
 This library is a lightweight modern wrapper around sqlite C api .
 
-**Heads up!**
-The current master contains a _breaking change_ compared to the 2.4 release for database names with non-ASCII characters. For details, see [issue 82](https://github.com/aminroosta/sqlite_modern_cpp/issues/82).
-
 ```c++
 #include<iostream>
 #include <sqlite_modern_cpp.h>
@@ -323,6 +320,12 @@ The usual way works for installing:
 ```
 
 Note, there's nothing to make, so you there's no need to run configure and you can simply point your compiler at the hdr/ directory.
+
+Breaking Changes
+----
+
+- Databases with non-ASCII characters in their names created with versions up to 2.4 are not found by the current master.
+You have to manually rename them to their actual (UTF-8 encoded) name.
 
 Package managers
 ----
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -133,7 +133,6 @@ namespace sqlite {
 
 		database_binder(database_binder&& other) :
 			_db(std::move(other._db)),
-			_sql(std::move(other._sql)),
 			_stmt(std::move(other._stmt)),
 			_inx(other._inx), execution_started(other.execution_started) { }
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -5,6 +5,9 @@ sqlite modern cpp wrapper
 
 This library is a lightweight modern wrapper around sqlite C api .
 
+**Heads up!**
+The current master contains a _breaking change_ compared to the 2.4 release for database names with non-ASCII characters. For details, see [issue 82](https://github.com/aminroosta/sqlite_modern_cpp/issues/82).
+
```
aminroosta: @zauguin Could you please add a **Breaking Changes** section at the end of REAME, i will update previous breaking changes and we will add new changes to that section :-)

Killili:
Hi.

I'm not sure why i did that u16string but i had to add that to get it working with Windows and Linux.
I think it was like that before and it did not work cross platform. It had something to do with wchar having different sizes. 

Sorry for the vague comment, its been a while ;)

zauguin:
A little history change revealed that before we used u16string, we used wstring. On linux this lead to a situation where we passed UTF-32 encoded data to function expecting UTF-16.
Do you mean this?

However that isn't related to this change, because this isn't about using wstring, but about passing UTF-8  data directly to SQLite without hiding it in the lower bits of UTF-16.

But I'll still test this on Windows before we merge it.

zauguin:
I tested this branch on Windows (Windows 10, Visual C++ 2015) without problems.
Especially UTF-8 and UTF-16 names work fine.
@Killili Is there something specific I should test?

Killili:
Sounds like the Bug i fixed back then. So i`m good with this change (because all my databases have English ASCII names ;). And it does not corrupt the contend.

PR #84, opened 2017-02-09
[RFC] Don't bundle SQLite

zauguin:
Why do we include the SQLite source code?
I couldn't find a place where it is used and it just becomes out-of-date.


aminroosta:
@zauguin I agree with you that it becomes out of date.
But it's easier to clone the repo and start changing the source code / running tests ... right away.

zauguin:
That's great, but wouldn't we have to change the Makefile to actually use the bundled SQLite?

aminroosta:
Ah, I thought we are using these files in the Makefile!
Truth to be told i didn't wrote the `./configure` stuff :-)

@zauguin Thanks for the PR, go ahead and merge it if that's the case 👍

zauguin:
Merged, if we ever need this, we can add it back together with the Makefile changes.

PR #86, opened 2017-02-13
Use std::uncaught_exceptions if possible

zauguin:
std::uncaught_exception will be deprecated with C++17 and might introduce weird bugs at edge cases.


aminroosta:
@zauguin This is really great, Thanks 😉
(we need to remain **modern** C++!)

PR #88, opened 2017-02-14
Include SQL statement in exception

### LINKED ISSUE #85: access to sql statement
Certainly for debugging and logging it would be very helpful to have access to the last sql statement either thru the database object or thru the exception handler ... like get_code() ... get_sql().

Is this possible?

zauguin:
Add the expanded query string to exceptions.
Additionally this makes the query accessible from the `database_binder`.

Resolves #85


zauguin:
I'm not sure why the test failed, it doesn't happen on my system.

zauguin:
Ok, it was a version problem.
@aminroosta Is there a general rule which SQLite versions we are compatible with?

aminroosta:
@zauguin The sqlite.c & sqlite.h files in the repo (that are removed now) where `v3.12.0`. 
Unless there is major breaking change in sqlite api, i suppose our wrapper should work with `3.x.y`.

PR #90, opened 2017-02-17
Fix build for modern standard libraries

zauguin:
The current build breaks, iff the standard library supports `<optional>`, but C++17 support isn't enabled.
Additionally we should not use _Reserved identifiers.
@aminroosta Should we make a bugfix release(3.0.1?) out of this?

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -9,12 +9,12 @@
 #include <vector>
 
 #ifdef __has_include
-#if __has_include(<optional>)
-#define _MODERN_SQLITE_STD_OPTIONAL_SUPPORT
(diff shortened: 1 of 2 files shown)


funilrys:
@zauguin it's the building process who fails or the new features?

zauguin:
It doesn't build independent of the used features. libc++ v4 and libstdc++v7 will be affected.

funilrys:
Okay that's clean now.
So I have to wait for @aminroosta's and gonna wait a bit before publishing update to AUR.
_________
If we don't have a bug fix release, I'm gonna push to AUR with a patch according to your PR @zauguin

aminroosta:
@funilrys I just released v3.1 with this PR.

PR #93, opened 2017-02-18
Alternative flags and encoding support

### LINKED ISSUE #71: open as readonly

zauguin:
Provides the functionality of #72 with the sqlite_config interface.

Fixes #71

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -406,14 +406,14 @@ namespace sqlite {
 	}
 	
 	enum class Encoding {
-	  AUTO = SQLITE_ANY,
+	  ANY = SQLITE_ANY,
 	  UTF8 = SQLITE_UTF8,
 	  UTF16 = SQLITE_UTF16
 	};
 	struct sqlite_config {
     int flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
     const char *gVfs = nullptr;
-    Encoding encoding = Encoding::AUTO;
+    Encoding encoding = Encoding::ANY;
 	};
 
 	class database {
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -421,26 +421,7 @@ namespace sqlite {
 		std::shared_ptr<sqlite3> _db;
 
 	public:
-		database(std::u16string const & db_name): _db(nullptr) {
-			sqlite3* tmp = nullptr;
-			auto ret = sqlite3_open16(db_name.data(), &tmp);
-			_db = std::shared_ptr<sqlite3>(tmp, [=](sqlite3* ptr) { sqlite3_close_v2(ptr); }); // this will close the connection eventually when no longer needed.
-			if(ret != SQLITE_OK) exceptions::throw_sqlite_error(ret);
-			//_db.reset(tmp, sqlite3_close); // alternative close. (faster?)
-		}
-
-		database(std::string const & db_name): _db(nullptr) {
-			sqlite3* tmp = nullptr;
-			auto ret = sqlite3_open(db_name.data(), &tmp);
-			_db = std::shared_ptr<sqlite3>(tmp, [=](sqlite3* ptr) { sqlite3_close_v2(ptr); }); // this will close the connection eventually when no longer needed.
-			if(ret != SQLITE_OK) exceptions::throw_sqlite_error(ret);
-			//_db.reset(tmp, sqlite3_close); // alternative close. (faster?)
-		}
-
-		database(std::shared_ptr<sqlite3> db):
-			_db(db) {}
-
-		database(const std::string &db_name, const sqlite_config &config): _db(nullptr) {
+		database(const std::string &db_name, const sqlite_config &config = {}): _db(nullptr) {
 			sqlite3* tmp = nullptr;
 			auto ret = sqlite3_open_v2(db_name.data(), &tmp, config.flags, config.gVfs);
 			_db = std::shared_ptr<sqlite3>(tmp, [=](sqlite3* ptr) { sqlite3_close_v2(ptr); }); // this will close the connection eventually when no longer needed.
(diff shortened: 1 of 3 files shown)


hdr/sqlite_modern_cpp.h
```diff
@@ -399,41 +399,44 @@ namespace sqlite {
 		);
 	}
 	
+	enum class Encoding {
+	  AUTO = SQLITE_ANY,
+	  UTF8 = SQLITE_UTF8,
+	  UTF16 = SQLITE_UTF16
+	};
```
aminroosta: @zauguin Use the same name as sqlite.
```C++
enum class Encoding {
  ANY = SQLITE_ANY,
  UTF8 = SQLITE_UTF8,
  UTF16 = SQLITE_UTF16
};
```

hdr/sqlite_modern_cpp.h
```diff
@@ -399,41 +399,44 @@ namespace sqlite {
...
 	struct sqlite_config {
+    int flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
```
aminroosta: @zauguin Use an enum for flags instead of `int`.
```c++
enum class Flags {
  OPEN_READONLY = SQLITE_OPEN_READONLY,
  OPEN_READWRITE = SQLITE_OPEN_READWRITE,
  OPEN_CREATE = SQLITE_OPEN_CREATE,
  OPEN_NOMUTEX = SQLITE_OPEN_NOMUTEX,
  OPEN_FULLMUTEX = SQLITE_OPEN_FULLMUTEX,
  OPEN_SHAREDCACHE = SQLITE_OPEN_SHAREDCACHE,
  OPEN_PRIVATECACH = SQLITE_OPEN_PRIVATECACH,
  OPEN_URI = SQLITE_OPEN_URI
};
```

hdr/sqlite_modern_cpp.h
```diff
@@ -399,41 +399,44 @@ namespace sqlite {
...
+    const char *gVfs = nullptr;
```
aminroosta: @zauguin rename this to `zVfs` same as sqlite documentation.

hdr/sqlite_modern_cpp.h
zauguin: For a `enum class` `operator |` isn't defined, so this leads to quite some boilerplate. Do need `operator&` too?

hdr/sqlite_modern_cpp.h
```diff
@@ -399,41 +399,44 @@ namespace sqlite {
...
+			auto ret = sqlite3_open_v2(db_name.data(), &tmp, config.flags, config.gVfs);
 			_db = std::shared_ptr<sqlite3>(tmp, [=](sqlite3* ptr) { sqlite3_close_v2(ptr); }); // this will close the connection eventually when no longer needed.
 			if(ret != SQLITE_OK) exceptions::throw_sqlite_error(ret);
-			//_db.reset(tmp, sqlite3_close); // alternative close. (faster?)
+			if(config.encoding == Encoding::UTF16)
+			  *this << R"(PRAGMA encoding = "UTF-16";)";
 		}
 
-		database(std::string const & db_name): _db(nullptr) {
+		database(const std::u16string &db_name, const sqlite_config &config = {}): _db(nullptr) {
```
aminroosta: Make the default encoding for this constructor `UTF16`, this changes the behavior we had in previous versions.
@zauguin

aminroosta:
@zauguin Looks very good to me, please add a unit test and update documentation.
Thanks for the great work you have done with these recent PRs 😉

hdr/sqlite_modern_cpp.h
aminroosta: @zauguin I didn't know that, Is there a workaround for this limitation?
No we don't need `operator &`.

hdr/sqlite_modern_cpp.h
zauguin: I don't really understand this comment. By default the encoding is ANY, so in line 434 we select encoding UTF16.

hdr/sqlite_modern_cpp.h
aminroosta: @zauguin I just tested a hackish way, let me know what you think
```c++
enum class test { A = 1, B = 2, C = 4 };

test operator|(const test& a, const test& b) {
  test ret;
  ret = static_cast<test>(static_cast<int>(a) | static_cast<int>(b));
  return ret;
};

int main() {
  test c = test::A | test::B;
  cout << static_cast<int>(c) << endl; // prints 3 as expected :-)
}
```

hdr/sqlite_modern_cpp.h
aminroosta: @zauguin You are right, my mistake :-)

hdr/sqlite_modern_cpp.h
zauguin: Thanks, I integrated it.

zauguin:
This will hit the same Visual C++ bug as #94.
To implement the work-around I would refactor the conversion into a separate method. Then we could make it a little bit more generic. Is there interest in `wstring ` and `u32string` support?

aminroosta:
@zauguin I thinks its better to keep it simple, `std::string` & `std::u16string` are enough :-)

zauguin:
I only add tests for read-only and tests, I don't know anything about the other ones, but there is no reason why they shouldn't work.

hdr/sqlite_modern_cpp.h
```diff
@@ -405,41 +405,61 @@ namespace sqlite {
...
+	enum class Flags {
+		OPEN_READONLY = SQLITE_OPEN_READONLY,
+		OPEN_READWRITE = SQLITE_OPEN_READWRITE,
+		OPEN_CREATE = SQLITE_OPEN_CREATE,
+		OPEN_NOMUTEX = SQLITE_OPEN_NOMUTEX,
+		OPEN_FULLMUTEX = SQLITE_OPEN_FULLMUTEX,
+		OPEN_SHAREDCACHE = SQLITE_OPEN_SHAREDCACHE,
+		OPEN_PRIVATECACH = SQLITE_OPEN_PRIVATECACHE,
+		OPEN_URI = SQLITE_OPEN_URI
+	};
```
zauguin: @aminroosta What do you think about omitting OPEN_? It looks a bit redundant. If you prefer to keep it we might move it to the class name: `OpenFlags::READONLY` looks better than `Flags::OPEN_READONLY`

aminroosta: @zauguin `OpenFlags::READONLY` looks better, I'm for it 😉 
BTW do you think we should make a new release after this PR?

zauguin: Right now I'm working on `std::variant` support and I would really like to see this in the next release, especially for supporting functions with different argument types, but apart from that, go for it.

PR #95, opened 2017-02-21
std::variant support

zauguin:
The motivating use case for this are custom functions. With `variant`s you can create functions with flexible argument/return types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,9 +342,8 @@ If your columns may have flexible types, you can use C++17's `std::variant` to e
 
 ```c++
 db << "CREATE TABLE tbl (id integer, data);";
-db << "INSERT INTO tbl VALUES (?, ?, ?);" << 1 << vector<int> { 1, 2, 3};
-unique_ptr<string> ptr_null; // you can even bind empty unique_ptr<T>
-db << "INSERT INTO tbl VALUES (?, ?, ?);" << 2 << 2.5;
+db << "INSERT INTO tbl VALUES (?, ?);" << 1 << vector<int> { 1, 2, 3};
+db << "INSERT INTO tbl VALUES (?, ?);" << 2 << 2.5;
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -336,6 +336,41 @@ You can enable boost support by defining _MODERN_SQLITE_BOOST_OPTIONAL_SUPPORT b
...
 
+Variant type support (C++17)
+----
+If your columns may have flexible types, you can use C++17's `std::variant` to extract the value.
+
+```c++
+db << "CREATE TABLE tbl (id integer, data);";
+db << "INSERT INTO tbl VALUES (?, ?, ?);" << 1 << vector<int> { 1, 2, 3};
+unique_ptr<string> ptr_null; // you can even bind empty unique_ptr<T>
+db << "INSERT INTO tbl VALUES (?, ?, ?);" << 2 << 2.5;
```
aminroosta: @zauguin Here It should be two question marks, right? `INSERT INTO tbl values (?, ?)`

README.md
```diff
@@ -336,6 +336,41 @@ You can enable boost support by defining _MODERN_SQLITE_BOOST_OPTIONAL_SUPPORT b
...
+		>> [](std::variant<vector<int>, double> data) {
+			if(data.index() != 1) {
+				cerr << "ERROR: we expected a blob" << std::endl;
+			}
+
+			for(auto i : get<vector<int>>(data)) cout << i << ","; cout << endl;
+		};
+
+db << "select age,name,img from tbl where id = 2"
+		>> [](std::variant<vector<int>, double> data) {
```
aminroosta: @zauguin `age, name, img` are not defined in the table and the lambda is only taking one argument.

aminroosta:
@zauguin Looks great to me ;-)

PR #96, opened 2017-02-22
[RFC] Initial test of database_binder_hooks and Query Cancel

Killili:
Hi,

this is my initial vision of an interface for having custom hooks into the inner workings of the library.

Its not complete yet i just posted it for review and to see what you guyes think about it. And if its a worthy addtion to the project or if it just makes it more complex for no real gain.

I have a little test case implemented that basicly just logs every sql_step with pre and post hook.
This might seem like a total nonfeature but it already showed a bug in the current master i will explain in another post. And would help people to log their SQL statments or have a custom mutex scheme around sqlite3_step.

```c++
class logging_hooks: public database_binder_hooks {
	virtual bool pre_step(const database_binder& binder) override {
		cout << "pre_step: " << binder.sql() << "\n"; // LogSQL
		return true;
	}

	virtual bool post_step(int sql_result) override {
		cout << "post_step " << sql_result << "\n"; // Log return code
		return sql_result == SQLITE_ROW;
	}


	virtual bool post_callback(bool user_return) override {
		cout << "post_callback " << user_return << "\n"; // Log Lambda response
		return user_return;
	}


	virtual void error_throw() override { // Custom error handling
		cout << "error_throw\n";
	}

};

sqlite_config config;
config.hooks = make_shared<logging_hooks>(); // Create logging class
database db(":memory:",config); // enable logging while creating
```

There is another addition to the standard way the Lib works in that youre lambda can return a bool now to break out of an select query processing. I needed that for example to exit a thread that processes a very long query gracefully.

```c++
try {
	db << "select * from test order by number" >> [](string text, int number) {
		cout << "1st " << text << "," << number << "\n";
		return number != 1; // break select loop on specific number
	};
} catch(sqlite::sqlite_exception& ex) {
	cout << "We know that we might not process all rows\n";
}
```

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -35,11 +35,11 @@ namespace sqlite {
 	public:
 		sqlite_exception(const char* msg, std::string sql, int code = -1): runtime_error(msg), code(code), sql(sql) {}
 		sqlite_exception(int code, std::string sql): runtime_error(sqlite3_errstr(code)), code(code), sql(sql) {}
-		int get_code() {return code;}
-    std::string get_sql() {return sql;}
+		int get_code() { return code; }
+		std::string get_sql() { return sql; }
 	private:
 		int code;
-    std::string sql;
+		std::string sql;
 	};
 
 	namespace exceptions {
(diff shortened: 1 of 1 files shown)


zauguin:
This looks similar to `sqlite3_trace_v2`, but on a _sqlite_modern_cpp_ level. So I would prefer to directly expose this functionality.
That said, I never tried to use this, so I can't say if this is a plausible alternative.

Killili:
Yes looks like my example use case could be made using that i choose logging as an obvious example, but this feature has a bit more functionality to it. Especially the capacity to cancel the processing of a query on every step without throwing exceptions from the callback.

PR #98, opened 2017-02-22
Fix 97

Killili:


Killili:
Hm, Travis is broken ... or is it my code?
```
$ sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"
Cannot add PPA: 'ppa:ubuntu-toolchain-r/test'.
Please check that the PPA name or format is correct.
The command "sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"" failed and exited with 1 during ```

File: tests/prepared_statment.cc [hunk 1]
```diff
@@ -75,6 +75,28 @@ int main() {
...
 		}
 
+		{
+			auto prep = db << "select ?";
+
+			prep << 5;
+
+			prep.execute();
+			try {
+				prep.execute();
```
zauguin: I think you want to add a `exit(EXIT_FAILURE);` here to handle the case that the error is not detected.

hdr/sqlite_modern_cpp.h
```diff
@@ -173,7 +174,12 @@ namespace sqlite {
       return sqlite3_sql(_stmt.get());
     }
 
-		void used(bool state) { execution_started = state; }
+		void used(bool state) {
+			if(execution_started == true && state == true) {
+				throw sqlite_exception("Already used statement executed again! Please reset() first!",sql());
```
zauguin: Maybe this is a `exception::misuse` or we should introduce a custom class?

Killili: I was pondering that too, but the "special" exceptions are the SQLite errors so i did not want to add custom ones there. Because SQLite wouldnt throw an error there.

tests/prepared_statment.cc
Killili: youre right

hdr/sqlite_modern_cpp.h
zauguin: Actually there already are custom ones (`no_rows`, `more_rows`) too.

Killili: i see, those are new to me... although i might have wanted them ;)

zauguin: Maybe in this situation `assert` would be a better fit, because this can only be triggered by coding errors.

Killili: Hm, i like to throw stuff and i would like to replace the whole exception throwing with a user replaceable callback later (secret objective of #96 i will explain there).
But i see the point in using assert here where it not for me calling this stuff from an embedded scripting language so this can be a runtime error for me.

Killili:
And Travis is happy 😀 are we happy, too?

PR #102, opened 2017-03-28
Fix exceptions so they can be caught by const reference

cbeiraod:
I am not sure if I am making the pull request to the right branch or if I have missed something, but the methods of the exception class on the main branch do not allow the exceptions to be caught by const reference as is normally recommended to do.

Modifying the methods to be const fixes the issue.


zauguin:
Thanks

PR #105, opened 2017-04-07
Support all integer types

### LINKED ISSUE #104: Support for long and long long missing?
Hi,
according to the documentation in your examples this library should support long and long long as types:

```
// inserts a new user record.
		// binds the fields to '?' .
		// note that only types allowed for bindings are :
		//      int ,long, long long, float, double
		//      string , u16string
		// sqlite3 only supports utf8 and utf16 strings, you should use std::string for utf8 and std::u16string for utf16.
		// note that u"my text" is a utf16 string literal of type char16_t * .
		db << "insert into user (age,name,weight) values (?,?,?);"
			<< 20
			<< u"bob"
			<< 83.25;
```

But when I'm trying use it with a long value I get this error:

```
 error: undefined reference to 'sqlite::database_binder& sqlite::operator<< <long>(sqlite::database_binder&, long const&)'

```

Casting the value to an int before using it in the statement makes it work. 
When I took a look at the `sqlite_modern_cpp.h` file I can't find a fitting function for long values.

So is the documentation wrong or am I just to stupid to use it with a long?

zauguin:
Fixes #104 and lifts the restrictionon on only `int`, `long` and `long long`.

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -285,8 +285,6 @@ namespace sqlite {
 #endif
 
 
-		template<typename T> friend database_binder& operator <<(database_binder& db, const T& val);
-		template<typename T> friend void get_col_from_db(database_binder& db, int inx, T& val);
 		/* for vector<T, A> support */
 		template<typename T, typename A> friend database_binder& operator <<(database_binder& db, const std::vector<T, A>& val);
 		template<typename T, typename A> friend void get_col_from_db(database_binder& db, int inx, std::vector<T, A>& val);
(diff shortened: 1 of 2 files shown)


hdr/sqlite_modern_cpp.h
```diff
@@ -848,6 +846,29 @@ namespace sqlite {
 	 inline void store_result_in_db(sqlite3_context* db, const std::u16string& val) {
 		 sqlite3_result_text16(db, val.data(), -1, SQLITE_TRANSIENT);
 	}
+
+	// Other integer types
+	 template<class Integral, class = typename std::enable_if<std::is_integral<Integral>::value>::type>
+	 inline database_binder& operator <<(database_binder& db, const Integral& val) {
+		return db << static_cast<sqlite3_int64>(val);
```
aminroosta: This will store `int` values as 64 bit values.
Please restrict the template with a condition like this. `is_integral<Integral> and sizeof(Integral) > sizeof(int)`.
And overload another template that stores `int` values and has condition like this `is_integral<Integral> and sizeof(integral) == sizeof(int)`

aminroosta: @zauguin

zauguin: Actually `int` will still be handled by the `int` overload. Your version would lead to a failure for `unsigned int`, if the value doesn't fit into an `int`. Additionally this is the original implementation of `sqlite3_bind_int`:

```c
SQLITE_API int sqlite3_bind_int(sqlite3_stmt *p, int i, int iValue){
  return sqlite3_bind_int64(p, i, (i64)iValue);
}
```
So we don't really have to care about calling it for small values anyway.

aminroosta: @zauguin aah, wonderful  👍

Issue #109, opened 2017-04-22
sqlite error log?


bstarynk:
It would be nice to be able to interface with [sqlite3_config](https://sqlite.org/c3ref/config.html) for [error logging](https://sqlite.org/errlog.html) with [SQLITE_CONFIG_LOG](https://sqlite.org/c3ref/c_config_covering_index_scan.html).

Perhaps we would want to be able to code

    sqlite::error_log([&](int errcode, const char*msg) {
      std::clog<<"SQLITE ERROR " << errcode << ":" << msg << std:endl;
    });

What do you think?

I might provide a small patch if we agree on that API.

Thanks.

[Basile Starynkevitch](http://starynkevitch.net/Basile/)  :   `basile`  at `starynkevitch` dot `net`
(France)

aminroosta:
@bstarynk I think we should somehow use our `sqlite::exceptions` namespace.
We don't want users figuring out `error_type` from `error_code`.  

A simple solution may be:
```c++
sqlite::error_log([](sqlite_exception e) {
     cerr  << e.get_code() << ": " << e.what() << " during "
	      << e.get_sql() << endl; // Now we also have .get_sql() :-)
});
```
I am also wondering how can we efficiently overload that function on exceptions from `sqlite::exceptions` namespace. @zauguin :-)

```c++
sqlite::error_log([](sqlite::exceptions::internal e) { });
sqlite::error_log([](sqlite::exceptions::full e) { });
// ...
```

zauguin:
I think it would be easier to specify all overloads in a sigle call, like
``` c++
sqlite::error_log(
    [](sqlite::exceptions::internal e) {
      // Do sth.
    }),
    [](sqlite::exceptions::full e) {
      // Do sth.
    }),
    ...
    );
```
Otherwise we would need some kind of runtime list for this.

This can easily be implemented on top of the basic functionality. Anyway `.get_sql()` will return an empty string, because AFAICT the current SQL statement is not available.

aminroosta:
@zauguin Looks perfect to me 👍 

@bstarynk What do you think, does zauguin api works for you?

bstarynk:
I am not sure to follow you. It looks you think of C++ exceptions.
I was just thinking of adding an error log function.

aminroosta:
@bstarynk 
```c++
sqlite::error_log([&](int errcode, const char*msg) {
[...]

PR #110, opened 2017-04-23
Extended errors

zauguin:
Expose extended result codes from SQLite.

The list was moved into a macro based separate file to avoid duplication the list, which seemed error-prone, especially since the extended code list might chnage more often.


aminroosta:
@zauguin This is just a suggestion, can you move `sqlite::exceptions` namespaces and all the macro stuff into another header (exceptions.h).

besides that as always great work, well done 👍, we just need a unit test and updated readme :-)

zauguin:
@aminroosta I moved it into a `errors.h` header und renamed the `exceptions` namespace to represent the more general usage of this classes in preparation for #109. The old named is preserved in form of an alias.
Maybe `sqlite_exception` should be renamed too?

aminroosta:
@zauguin renaming `sqlite_exception` will be backwards incompatible, so let's just keep it :-)

This PR is perfect to me, the readme & tests are all great 👍 
I'm merging it, we will have another PR for #109

PR #111, opened 2017-04-25
Error logging

### LINKED ISSUE #109: sqlite error log?

zauguin:
Implement SQL error logging based on error classes.

Closes #109.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,7 +412,7 @@ The call to `sqlite::error_log` has to be the first call to any `sqlite_modern_c
 			cerr  << e.get_code() << ": " << e.what() << endl;
 		},
 		[&](errors::misuse& e) {
-			/* You can behave differently to specific exceptions */
+			/* You can behave differently to specific errors */
 		}
 	);
 	database db(":memory:");
--- a/README.md
+++ b/README.md
@@ -402,6 +402,30 @@ Additionally you can use `get_sql()` to see the SQL statement leading to the err
 	   catch(sqlite::errors::constraint_primarykey e) {  } */
 ```
 
+You can also register a error logging function with `sqlite::error_log`.
+The `<sqlite_modern_cpp/log.h>` header has to be included to make this function available.
+The call to `sqlite::error_log` has to be the first call to any `sqlite_modern_cpp` function by your program.
+
+```c++
+	error_log(
+		[&](sqlite_exception& e) {
+			cerr  << e.get_code() << ": " << e.what() << endl;
+		},
+		[&](errors::misuse& e) {
+			/* You can behave differently to specific exceptions */
+		}
+	);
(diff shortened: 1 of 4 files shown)


zauguin:
@bstarynk Maybe you can help testing this?

aminroosta:
@bstarynk Could you please comment :-)

bstarynk:
I am not sure to understand how this should be used in practice. Maybe more comments could be useful. I am not sure to only want the error logging function to be activated *after* throwing some exception (but perhaps *before* that).

BTW, in my [melt-monitor](https://github.com/bstarynk/melt-monitor) (on github, e.g. commit [5b29edf1ef1b6621a0...](https://github.com/bstarynk/melt-monitor/commit/5b29edf1ef1b6621a0c0309fab71de8df07ffa1a)...) I have an additional wish. I am using extensively Ian Taylor's [libbacktrace](https://github.com/ianlancetaylor/libbacktrace), and I would like to print the backtrace *before* throwing any sqlite exception.

bstarynk:
BTW, I would also like such new features to be testable with preprocessor flags, that is have perhaps `#define SQLITE_MODERNCPP_HAS_ERRORLOG 1` so that my application could code `#if SQLITE_MODERNCPP_HAS_ERRORLOG` ....

zauguin:
@bstarynk This may be a documentation issue, why do you think the error logging is called after throwing the exception?

The error logging, based on the SQLite log functionality, is kind of independant of C++ exceptions. During execution, if SQLite detects an error, it calls the error log callback. Your handler is called with a `sqlite_exception` derived parameter to encode the error code, but this isn't a thrown exception. The exception is independently thrown at a later point.

So using libbacktrace should work, just replace the `cerr << ...` by your libbacktrace based error handling functionality.

zauguin:
This feature is implemented in a separate header, so with most modern compiler (gcc >= 5, clang, VC++ 2015(?)) you can test for it with [`__has_include`](wg21.link/p0061). For example

```c++
#if __has_include(<sqlite_modern_cpp/log.h>)
...
#endif
```

PR #113, opened 2017-04-28
adding sqlite::endr, and ability to give functions to database_binder

bstarynk:
We add the ability to force the end of a prepared statement with `endr` and the ability to give functions to them.

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -91,6 +91,10 @@ namespace sqlite {
 			}
 			
 		}
+		void endr() {
+		  execute();
+		  reset();
+		}
 		
 		std::string sql() {
 #if SQLITE_VERSION_NUMBER >= 3014000
@@ -285,6 +289,10 @@ namespace sqlite {
 			});
 		}
 	};
+
+	inline void endr(database_binder& db) {
+	  db.endr();
+	}
 
 	namespace sql_function_binder {
 		template<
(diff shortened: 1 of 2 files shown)


zauguin:
@bstarynk Have you seen the `++` operator?
Your `ppsi << "zero" << 0 << endr;` can be written today as `(ppsi << "zero" << 0)++;`

tests/prepared_statment.cc
```diff
@@ -99,6 +99,11 @@ int main() {
 
 		}
 
+		auto ppsc = db << "CREATE TABLE tnums(name VARCHAR(30) PRIMARY KEY ASC NOT NULL UNIQUE, num INT NOT NULL);";
```
zauguin: This statement is never executed. Either don't save the prepared statement to execute it implicitly or call `execute`/`endr`/`operator++`.

hdr/sqlite_modern_cpp.h
```diff
@@ -91,6 +91,10 @@ namespace sqlite {
 			}
 			
 		}
+		void endr() {
+		  execute();
+		  reset();
+		}
```
zauguin: We already have the operator, so I don't think we need a member function. If we keep this function, we may prefer a more descriptive name.

hdr/sqlite_modern_cpp.h
```diff
@@ -787,6 +795,24 @@ namespace sqlite {
...
+	   fun(&db);
+	   return db;
+	 }
+
+	 /// apply a lambda
+	 inline database_binder& operator<< (database_binder& db, std::function<void(database_binder&)>f) {
+	   f(db);
+	   return db;
+	 }
+
```
zauguin: Can we use our `function_traits` instead of `std::function` here to avoid type-erasure?
What is the use-case for calling this with a lambda?

PR #120, opened 2017-05-14
Auto reset after statement execution

zauguin:
Deprecate `reset()` and reset the statement after every execution to avoid bugs.
The patch changes the behaviour of `operator++` to no longer set `used(false)`.
This fixes the loop example in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,13 +117,10 @@ ps << tmp;
 // But beware that it will execute on destruction if it wasn't executed!
 ps >> [&](int a,int b){ ... };
 
-// after a successfull execution the statment needs to be reset to be execute again. This will reset the bound values too!
-ps.reset();
-
+// after a successfull execution the statment an be executed again, but the bound values are resetted.
 // If you dont need the returned values you can execute it like this
-ps.execute(); // the statment will not be reset!
-
-// there is a convinience operator to execute and reset in one go
+ps.execute();
+// or like this
 ps++;
 
 // To disable the execution of a statment when it goes out of scope and wasn't used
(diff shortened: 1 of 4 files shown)


Killili:
Hi,
im not sure why i split that in two. The only thing i can image is that i wanted to prevent resetting in case its not needed, for performance sake. I remember talking to @aminroosta about that split, too.

But anyway if i understand the change correct, then now it will reset the statement even if its not used as prepared statment? So there is always  create->execute->clear->reset->finalize even if clear and reset are not need like in 99% of all use cases i guess?

aminroosta:
@zauguin I think just remove the `reset` function, I'm thinking the next version should be v4.
I'm also thinking about removing the `boost` stuff.

README.md
```diff
@@ -117,13 +117,10 @@ ps << tmp;
 // But beware that it will execute on destruction if it wasn't executed!
 ps >> [&](int a,int b){ ... };
 
-// after a successfull execution the statment needs to be reset to be execute again. This will reset the bound values too!
-ps.reset();
-
+// after a successfull execution the statment an be executed again, but the bound values are resetted.
```
aminroosta: @zauguin little type ... the statement **C**an be executed ...

Killili:
Nice, now i like it. Thx for fixing my inconsistencys  ;)

zauguin:
@aminroosta I removed `reset` now. Is there any place for documenting breaking changes in the `master` branch or do we just have to remember it for the next release?

aminroosta:
@zauguin Yes, we just need to document it in the next release :-)

PR #125, opened 2017-05-23
blob_t implementation

aminroosta:
@zauguin @Killili Please review this PR :-)
There might be a better approach, so far i've came up with this.

#122

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -62,6 +62,39 @@ namespace sqlite {
 		static void iterate(Tuple&, database_binder&) {}
 	};
 
+	template<typename>
+	struct blob_t;
+
+	template<typename T, typename A>
+	struct blob_t<std::vector<T,A>> {
+	  blob_t() { }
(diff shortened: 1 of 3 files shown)


zauguin:
Except for the `sqlite_config`, this is with `.vec` the only place the interface relies on public data members. What do you think about using sth. like `operator *`, because `blob_t` is a wrapper and we want to access the underlying object?

Killili:
I like it, it makes a lot more sense than just having ```vector<>``` as an option for ```operator<<``` to make blobs.
What this implies is "use a blob to store x" and x is a vector because thats all we can handle for now. But ```blob()``` can be expanded on in the future to handle all kinds of stuff.
So i do not like that it is an interface for vector only but i like it for the expend-ability later on. It sets an example for how we might handle custom data types later on or even explicitly choosing the SQL Type like:
```c++
auto val = get_something_numerical_maybe_float_or_int();
db << "insert ...." << val; // might end up numeric or real
db << "insert ...." << real(val); // will always be a real or error if for example its out of range
```

PR #126, opened 2017-05-24
Alternative iterator interface

zauguin:
I think this feels more natural than the interface from #103 and it makes lifetime questions easier.
The `database_binder` behaves like a container now:

```c++
for(auto &&row : db << "select age,name,weight from user where age > ? ;" << 21) {
  int _age;
  string _name;
  double _weight;
  row >> _age >> _name >> _weight;
  cout << _age << ' ' << _name << ' ' << _weight << endl;
}
// or
for(std::tuple<int, string, double> row : db << "select age,name,weight from user where age > ? ;" << 21) {
  cout << std::get<int>(row) << ' ' << std::get<string>(row) << ' ' << std::get<double>(row) << endl;
}
```
@Killili What do you think? It's not our usual stream-op interface, but I think this makes it clearer that the range doesn't exists independent from the statement.

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -239,7 +239,7 @@ namespace sqlite {
 #endif
 #ifdef MODERN_SQLITE_STD_OPTIONAL_SUPPORT
 		template <typename T>
-		struct is_sqlite_value< std::optional<T> > : public is_sqlite_value<T> {};
+		struct is_sqlite_value< optional<T> > : public is_sqlite_value<T> {};
 #endif
 
 #ifdef _MODERN_SQLITE_BOOST_OPTIONAL_SUPPORT
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -154,37 +154,9 @@ namespace sqlite {
 			used(true);
 		}
 
-		void _extract(std::function<void(void)> call_back) {
-			int hresult;
-			_start_execute();
+		void _extract(std::function<void(void)> call_back);
 
-			while((hresult = sqlite3_step(_stmt.get())) == SQLITE_ROW) {
-				call_back();
-			}
-
-			if(hresult != SQLITE_DONE) {
-				errors::throw_sqlite_error(hresult, sql());
-			}
-		}
-
-		void _extract_single_value(std::function<void(void)> call_back) {
-			int hresult;
-			_start_execute();
-
-			if((hresult = sqlite3_step(_stmt.get())) == SQLITE_ROW) {
-				call_back();
-			} else if(hresult == SQLITE_DONE) {
-				throw errors::no_rows("no rows to extract: exactly 1 row expected", sql(), SQLITE_DONE);
-			}
-
-			if((hresult = sqlite3_step(_stmt.get())) == SQLITE_ROW) {
-				throw errors::more_rows("not all rows extracted", sql(), SQLITE_ROW);
-			}
-
-			if(hresult != SQLITE_DONE) {
-				errors::throw_sqlite_error(hresult, sql());
-			}
-		}
+		void _extract_single_value(std::function<void(void)> call_back);
 
 #ifdef _MSC_VER
 		sqlite3_stmt* _prepare(const std::u16string& sql) {
@@ -396,6 +368,21 @@ namespace sqlite {
 
 	inline row_iterator database_binder::end() {
 		return row_iterator();
+	}
+	void database_binder::_extract(std::function<void(void)> call_back) {
+		for(auto &&row[[maybe_unused]] : *this)
+			call_back();
+	}
+
+	void database_binder::_extract_single_value(std::function<void(void)> call_back) {
+		auto iter = begin();
+		if(iter == end())
+			throw errors::no_rows("no rows to extract: exactly 1 row expected", sql(), SQLITE_DONE);
+
+		call_back();
+
+		if(++iter != end())
+			throw errors::more_rows("not all rows extracted", sql(), SQLITE_ROW);
 	}
 
 	namespace sql_function_binder {
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -360,7 +360,7 @@ namespace sqlite {
 				throw errors::more_rows("not all rows extracted", binder.sql(), SQLITE_ROW);
 		}
 	}
-	void database_binder::execute() {
+	inline void database_binder::execute() {
 		for(auto &&row : *this)
 			(void)row;
 	}
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -269,12 +269,6 @@ namespace sqlite {
 				get_col_from_db(*_binder, next_index++, result);
 				return *this;
 			}
-			template<class T, typename = typename std::enable_if<detail::is_sqlite_value<T>::value, value_type &>::type>
-			operator T() {
-				T result;
-				*this >> result;
-				return result;
-			}
 			template<class ...Types>
 			value_type &operator >>(std::tuple<Types...>& values) {
 				assert(!next_index);
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -306,8 +306,6 @@ namespace sqlite {
 			return *this;
 		}
 
-		// Not well-defined
-		row_iterator operator++(int);
 		friend inline bool operator ==(const row_iterator &a, const row_iterator &b) {
 			return a._binder == b._binder;
 		}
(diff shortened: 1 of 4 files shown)


Killili:
I like it more then #103 it looks more consistent with the rest of the API.

But now that i think about it, isnt it a lot like the prepared statement stuff?
```c++
auto prep = db << "select... where ?"; // create
prep << 21; // bind
auto itr = prep.begin() // "execute" // till here the prep statement does the same
auto row = itr++; // iterate
```
Maybe there is some synergy that could be used?

zauguin:
I added some documentation and rebased it on the current master.
The old code for executing prepared statements is now internally implemented by using this interface.

If noone objects I would merge it soon.

Killili:
I'm totaly out the loop here. Didnt touch C++ for the last 10+ Month. So this look like magic now ;)
So i can only rely on my past selfs comment up there and he liked it. So great work bringing it to this point. But someone else should take a look. @aminroosta

hdr/sqlite_modern_cpp.h
```diff
@@ -288,31 +209,193 @@ namespace sqlite {
...
+			}
+			template<class ...Types>
+			value_type &operator >>(std::tuple<Types...>& values);
+			template<class ...Types>
+			value_type &operator >>(std::tuple<Types...>&& values) {
+				return *this >> values;
+			}
+			template<class ...Types>
+			operator std::tuple<Types...>() {
+				std::tuple<Types...> value;
```
aminroosta: @zauguin This constructs an empty `tuple<Types...>` and then assigns to each tuple elements.
Can we avoid the overhead?

hdr/sqlite_modern_cpp.h
```diff
@@ -288,31 +209,193 @@ namespace sqlite {
...
+			return *this;
 		}
 
-		template<typename... Types>
-		void operator>>(std::tuple<Types...>&& values) {
-			this->_extract_single_value([&values, this] {
-				tuple_iterate<std::tuple<Types...>>::iterate(values, *this);
-			});
+		// Not well-defined
+		row_iterator operator++(int);
```
aminroosta: @zauguin I think we should not declare the postfix operator, a compile error is better than a linker one.

hdr/sqlite_modern_cpp.h
```diff
@@ -288,31 +209,193 @@ namespace sqlite {
...
+			return !(a==b);
 		}
 
-		template <typename Function>
-		typename std::enable_if<!is_sqlite_value<Function>::value, void>::type operator>>(
-			Function&& func) {
-			typedef utility::function_traits<Function> traits;
+	private:
+		database_binder *_binder = nullptr;
+		mutable value_type value{_binder}; // mutable, because `changing` the value is just reading it
```
aminroosta: @zauguin We have no const member function assigning to `value`, why do we need the `mutable` ?

aminroosta:
@zauguin Thanks for the really great work, This PR looks really interesting.
Sorry for the late reply, I missed the github notification.

I've added a review, Bu frankly it looks to me, good enough to merge at its current shape.

hdr/sqlite_modern_cpp.h
zauguin: If this is not mutable, operator* has to return a const reference  to `value_type` instead.
But reading from the `value_type` requires increasing the index for the next column, `next_index`.

I see four ways to solve this:

1. Make `operator *` and `operator ->` not const.
    This breaks requirements for iterators and might break code. IMO it would also be weird, because these functions are not changing anything.
2. Make `reference` the same as `value_type` and return a new `value_type` at every call to the operators.
    If this is ever used with a not range for loop, everytime `*iter` is used the read restarts at the first column.
    So this could lead to very hard to find bugs.
3. Make the `value_type` member functions `const` and `value_type::next_index` mutable. Then a const reference can be returned.
    This would work, but then we work `const` qualified instances of value_type, which don't act like `const` objects at all because they change if any function on them is called.
4. The current approach: A mutable `value`.
    This is certainly not perfect, but it moves the problems of 3. to the iterator type which is less visible to the user.

hdr/sqlite_modern_cpp.h
zauguin: The problem is that our interface always gets a referrence to a preexisting object.
So fixing this would require to rewrite `get_col_from_db` for every type to return the value instead of storing it in a parameter.
This is a streight forward conversion but I think it would be better to do this in a separate Pull Request because it would add another 100+ line of changes, so reviewing this together could easiliy become be a pain.

Also the overhead should be small or nonexistent: The values are default constructed, then in a probably inlined function a value is assigned, so most optimizers should eliminate the additional construction anyway.

hdr/sqlite_modern_cpp.h
zauguin: @aminroosta This was added to be compatible with the Ranges TS.
The `operator++` will normally not be necessary, but it is checked for with concepts.
But this was a hack in the first place and I agree moving a compile-time to a link-time error is bad.

We can either just take it out. This is the nice way and we don't lie about our type. The downside would themissing compatibility with Ranges-v3 or the Ranges TS.

We could also rewrite it to fail at compile time iff we are in an evaluated context, so using it would lead to a compile-time error, but checking for it would show it's existance.
Obviously this would really be a ugly hack, but we could interact with most libraries expecting "real iterators".

hdr/sqlite_modern_cpp.h
aminroosta: @zauguin Thanks for the clarification, Since `value_type` does not own the underlying data, I too think solution No 4 is the right choice here.

Right now `value_type` will only read data if dereferenced (`*value`) or when it is implicitly converted to `std::tuple`, basically a non-owing type.

What are your thoughts if we make it own the entire row?
`value_type` can own an `std::tuple` member, or drive from `std::tuple<Types...>`
We will initialize & update the underlying tuple in `row_iterator::operator++`.

aminroosta:
@zauguin A quick discussion about the branches we are using.

I will propose, we should have a `dev` branch,
This PR, also the PR I had for `blob_t` could go to that branch.
It will effectively be the branch for the next release.
We will mention it in the master readme.

This way can have an experimental(`dev`) & stable(`master`) branch.

hdr/sqlite_modern_cpp.h
aminroosta: I greed :-) 👍

hdr/sqlite_modern_cpp.h
aminroosta: > We can either just take it out. This is the nice way and we don't lie about our type. The downside would themissing compatibility with Ranges-v3 or the Ranges TS.

I think we should do that, unless we can come up with so called "real iterators" :-)

hdr/sqlite_modern_cpp.h
aminroosta: @zauguin If we drive from an `std::tuple`, can we take advantage of C++17 structured bindings?

zauguin: @aminroosta This sounds like the interface of #103.
The fundamental problem is determining the type and number of columns at compile time. If the value_type should own the values, the value_type has to know about <Types...>, so the types have to be passed as explicit template arguments.

zauguin:
@aminroosta This sounds like a great plan. Will you create the branch and set everything up?

aminroosta:
@zauguin Great, I will try to setup that by the end of this week.

I am planning to do the following: 
- [ ] Fix broken travis builds, happened after migrating to an organization.
- [ ] Remove boost support.
- [x] Replace autotools with Cmake & add Catch.
- [ ] Migrate our tests to use Catch.

aminroosta:
Here is the WIP https://github.com/SqliteModernCpp/sqlite_modern_cpp/compare/master...aminroosta:dev?expand=1

aminroosta:
@Killili 
I have removed my old branches from the repo.
(except blob_t, which I will remove after merging #125).

There are 3 branches from you too ([fix-138](https://github.com/SqliteModernCpp/sqlite_modern_cpp/tree/Fix-%2338), [nulltype](https://github.com/SqliteModernCpp/sqlite_modern_cpp/tree/nulltype), [custom_db](https://github.com/SqliteModernCpp/sqlite_modern_cpp/tree/custom_db)), could you please have a look.
Is it ok if we remove them? since we want to only keep master & dev from now on.

hdr/sqlite_modern_cpp.h
zauguin: @aminroosta How do you prefere to move forward here? Do you prefer the interface opions discussed in #103 or should we continue with this one?

Killili:
@aminroosta i'm not sure what these branches do, so they are probably not important.

aminroosta:
@Killili thanks, I deleted all of them.

@zauguin There is a merge conflict, please resolve this so we can merge this PR.
Thanks.

PR #132, opened 2017-06-14
Allow whitespace after semicolons

### LINKED ISSUE #131: Newline after semicolon is no longer supported since 3.2
The following code works perfectly with version 3.1 but is not supported anymore from 3.2.

```
	database db(":memory:");
	db<<"CREATE TABLE user (id INTEGER);\n";
```

Its happening because of the newline at the end of the statement.

**Since people write more readable codes and directly copy paste sql from sql-console,  this newline is very much important.**
E.g. 

```
const char* createUserTableSql = R"sqlcode(
CREATE TABLE IF NOT EXISTS user (
	_id integer primary key autoincrement not null,
	email text not null unique,
	password text not null
);
)sqlcode";
```

zauguin:
Fix #131.

Similar problems may arise with comments after the semicolon, but I don't know how to fix this without effectively writing a parser for SQLite comments.

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -180,7 +180,7 @@ namespace sqlite {
 			const char *remaining;
 			hresult = sqlite3_prepare_v2(_db.get(), sql.data(), -1, &tmp, &remaining);
 			if(hresult != SQLITE_OK) errors::throw_sqlite_error(hresult, sql);
-			if(!std::all_of(remaining, sql.data() + sql.size(), [](char ch) {return std::isblank(ch);}))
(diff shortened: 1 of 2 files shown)


aminroosta:
@zauguin 

I can think of two options
1- drop support for `Multiple semicolon separated statements are unsupported` and explain it in readme.
2- merge this PR and explain `comments after ;` in readme.
3- write a parser!

I think We would be better of not to preprocess sql statements :-)

zauguin:
@aminroosta Do you prefer 1. or 2.? Personally I like 2 more, because comments after ; are not that useful from C++ to begin with and lead to a error message, while in option 1 the error occurs without any notification.

aminroosta:
> comments after ; are not that useful from C++ to begin with and lead to a error message

@zauguin Makes perfect sense, lets go with 2 then 👍

PR #135, opened 2017-10-17
Adding std::experimental::optional

cbeiraod:
On macOS, despite the compiler already supporting std::optional, it seems the optional library does not yet exist, meaning we have to rely on std::experimental optional.

I have added that possibility by adding a check for `_MODERN_SQLITE_EXPERIMENTAL_OPTIONAL_SUPPORT`, similar to what is done for the boost::optional. It was implemented in such a way that the experimental::optional is only used if explicitly requested and available.

I also consolidated the implementation of the different optional types. apart from belonging to different namespaces, all optional types work the same, so they were merged together, using the ```using optional = std::optional<T>``` idiom.

The changes should not break existing code which depends on the previous behaviour, but I do not have the normal optional library nor boost installed, so I can not test this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,7 +325,7 @@ int main() {
 
 If you do not have C++17 support, you can use boost optional instead by defining `_MODERN_SQLITE_BOOST_OPTIONAL_SUPPORT` before importing the `sqlite_modern_cpp` header.
 
-If your compiler does support C++17, but does not have optional (macOS for instance, at the time this was written), you can use std::experimental::optional instead by defining `_MODERN_SQLITE_EXPERIMENTAL_OPTIONAL_SUPPORT` before importing the `sqlite_modern_cpp` header.
+If the optional library is not available, the experimental/optional one will be used instead.
 
 **Note: boost support is deprecated and will be removed in future versions.**
 
--- a/README.md
+++ b/README.md
@@ -325,6 +325,8 @@ int main() {
 
 If you do not have C++17 support, you can use boost optional instead by defining `_MODERN_SQLITE_BOOST_OPTIONAL_SUPPORT` before importing the `sqlite_modern_cpp` header.
 
+If your compiler does support C++17, but does not have optional (macOS for instance, at the time this was written), you can use std::experimental::optional instead by defining `_MODERN_SQLITE_EXPERIMENTAL_OPTIONAL_SUPPORT` before importing the `sqlite_modern_cpp` header.
+
 **Note: boost support is deprecated and will be removed in future versions.**
 
 Variant type support (C++17)
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -26,20 +26,13 @@
 #endif
 
 #ifdef MODERN_SQLITE_STD_OPTIONAL_SUPPORT
-  #include <optional>
-  template<class T>
-  using optional = std::optional<T>;
-  #define MODERN_SQLITE_OPTIONAL_SUPPORT
+#include <optional>
 #elif _MODERN_SQLITE_EXPERIMENTAL_OPTIONAL_SUPPORT && __has_include(<experimental/optional>)
-  #include <experimental/optional>
-  template<class T>
-  using optional = std::experimental::optional<T>;
-  #define MODERN_SQLITE_OPTIONAL_SUPPORT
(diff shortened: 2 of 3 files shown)


aminroosta:
@cbeiraod Thank you for the PR, generally i would prefer not to add new macros to the library.
Right now @zauguin implementation uses `__cplusplus > 201404` to check for c++17 support.

```cpp
#ifdef __has_include
#if __cplusplus > 201402 && __has_include(<optional>)
#define MODERN_SQLITE_STD_OPTIONAL_SUPPORT
#endif
#endif
```

Could you please check your `__cplusplus` version?
I think something like this should work.
And we can eliminate the need for the extra macro.
```c++
#if (__cplusplus > 201402 && __has_include(<optional>) || (__cplusplus > YOUR_VERSION && __has_include(<experimental/optional>))
```

The same thing could be done for  `<experimental/variant>` too.

Thanks :-)

zauguin:
AFAICT the patch already checks for __cplusplus and __has_include(<experimental/optional>), so I think you could just remove `_MODERN_SQLITE_EXPERIMENTAL_OPTIONAL_SUPPORT`.

aminroosta:
@zauguin Can we use `__has_include` in c++14 compilers?

zauguin:
It depends. `__has_include` is supported in all versions of clang and since GCC version 5 independent of the selected language standard (>=C++11).
On the other hand Visual Studio didn't support it until the last release, but there was at least one version where `__has_include` was defined but broken. I think in this version `__has_include()` was always false.

So it should be safe to use, as long as we keep the `#ifdef __has_include` test.

If you think about dropping the __cplusplus check, at least for <optional> this doesn't work, because the actual <optional> header requires C++17. So in C++14 mode, the <optional> header is detected, but if it is included it breaks the build.

cbeiraod:
Hi,

I have checked the `__cplusplus` version (I am compiling on both my mac and a docker image) and on macOS it is defined as 201406, while on the docker image it is defined as 201500. So on both, C++17 is supposedly supported, but the optional library is not there. Only under experimental.

Note that I am still on macOS 10.12 and I have not upgraded to the recently released macOS 10.13, so perhaps on the newer OS the optional library is already available.

The reason I included the extra macro (`_MODERN_SQLITE_EXPERIMENTAL_OPTIONAL_SUPPORT`) was because I thought it would not make sense to include the experimental library by default. As such, I made it require an explicit request from the user to use the experimental library. This check can be easily removed if you feel it is not needed.

Despite this fix, which seems to work under the project where I am using the slite_modern_cpp library, I am getting some hard to debug errors. It seems that when using an optional variable, that instead of using the << operator defined for optional, it is picking up the regular << operator which then results in a compilation error. Perhaps someone has had experience with using optional before?

I noticed the test for optional is disabled on Travis, perhaps the error above is why?

zauguin:
Hi,

including the experimental library by default appears odd, but in my opinion it's the former name of a now stable library, so as long as there are no problems with unconforming implementations it should be fine.

The errors sound interesting. Does the test work for you? Or do you have an example?
Regarding the test on Travis, it's not disabled because it doesn't work (it works on my system), it's disabled because the tests run with C++14, so the <optional> library doesn't exists.

cbeiraod:
So, I tried the `make test`command, and the tests do not compile on macOS. The last message given, and the only message when I try to re-run `make test` is:

```
make: *** No rule to make target `endian.h', needed by `tests/flags.test'.  Stop.
```

I have tried to explicitly compile the std_optional.cc test with the following command:

```
g++ -g -Wall -Wextra -pedantic -W -O3 -Werror -std=c++14 -I ../hdr -lsqlite3 std_optional.cc
```

Which revealed a problem with the experimental::optional implementation, which does not have a reset method defined (used [here](https://github.com/aminroosta/sqlite_modern_cpp/blob/3c43aa3cbb7185eadfb3bd36c4de95f6a6d3f5a8/hdr/sqlite_modern_cpp.h#L809) and [here](https://github.com/aminroosta/sqlite_modern_cpp/blob/3c43aa3cbb7185eadfb3bd36c4de95f6a6d3f5a8/hdr/sqlite_modern_cpp.h#L818)). By modifying it to use nullopt instead of reset (only for experimental::optional), the test compiles without any problem and I am able to run it.

This fix seems to have fixed the problem within my application as well. So everything is working fine now, as far as I can tell.

Issue #137, opened 2017-11-05
Does not link on macOS El Capitan


crapp:
Hi,

thanks for this great project and all your work. Started using your library a week ago and I am pretty happy. Unfortunately I can not use it with sqlite3 provided by macOS El Capitan because of linker errors. 
```shell
$ /usr/bin/sqlite3 --version            
3.8.10.2 2015-05-20 18:17:19 2ef4f3a5b1d1d0c4338f8243d40a2452cc1f7fe4
```
Running your tests
```shell
$ make test
g++ -g -O2 -Wall -Wextra -pedantic -W -O3 -Werror -std=c++14 -I hdr/ tests/blob_example.cc -o tests/blob_example.test -I . -lsqlite3  ||\
        { \
          echo "echo 'Compile error!' ; return 126" > tests/blob_example.test ; \
          chmod +x tests/blob_example.test; \
        }
Undefined symbols for architecture x86_64:
  "_sqlite3_expanded_sql", referenced from:
      sqlite::database_binder::sql[abi:cxx11]() in ccYCHJW3.o
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
```
The function ```sqlite3_expanded_sql```was introduced on 2016-08-08 with sqlite 3.14. I am not sure but I would say macOS Sierra might also be affected. The latest iteration of Apples operating system (High Sierra) comes with a more recent version of sqlite3 so you should be fine there

I fixed the problem for my application by installing sqlite3 from macports. Nevertheless this is something you should definitively mention in the README file. 

There is an additional problem with your tests as you try to include the header ```endian.h```

```shell
g++ -g -O2 -Wall -Wextra -pedantic -W -O3 -Werror -std=c++14 -I hdr/ tests/flags.cc -o tests/flags.test -I . -lsqlite3  ||\
        { \
          echo "echo 'Compile error!' ; return 126" > tests/flags.test ; \
          chmod +x tests/flags.test; \
        }
tests/flags.cc:5:20: fatal error: endian.h: No such file or directory
 #include <endian.h>
                    ^
compilation terminated.
``` 
On macOS ```endian.h```resides in ```/usr/include/machine```.
How about this alternative

```c++
#include <sys/types.h>
```
This will take care of the correct includes on macOS and Linux

zauguin:
The library is supposed to detect sqlite 3.14 at compile-time. Have you installed the sqlite header separately from the sqlite library?
[...]

PR #138, opened 2017-11-05
Fix endian detection in flags test for macOS

zauguin:
Fixes the second part of #137.

Thanks to @crapp.


zauguin:
@crapp Can you test this on macOS?

crapp:
Yes this works. Thank you very much

PR #139, opened 2017-11-09
update the readme file

Chris-Bond:
Fixed indentation for Arch Linux package maintainer and added vcpkg to the list of package managers.


aminroosta:
@Chris-Bond It would be awesome if you could maintain a `vcpkg` package for the library! 😄

PR #144, opened 2018-01-22
Function to return the count of database changes.

RenathoAzevedo:


hdr/sqlite_modern_cpp.h
```diff
@@ -460,6 +460,10 @@ namespace sqlite {
 			return sqlite3_last_insert_rowid(_db.get());
 		}
 
+		sqlite3_int64 rows_modified() const {
```
zauguin: I think this should return the same type as `sqlite3_changes`, so `int` instead of `sqlite3_int64`.

RenathoAzevedo: Oh that's right, It should be a int. Thank you. By the way which IDE do you use to the project ?

zauguin: Personally I'm using [(Neo)vim](https://neovim.io/), so I don't use any IDE.
The other guys might use something different, AFAICT there is no standard environment for the project.

RenathoAzevedo: Nice, I'm used to use netbeans on cross compile development. on local machine I'm using atom editor. I used atom to edit the project , but I don't know why the identation and whitespaces  inconsistencies happened.

zauguin:
@aminroosta I think this should go to the `dev` branch too, right?

aminroosta:
@zauguin Yea, this will go to `dev` branch.

PR #145, opened 2018-02-03
dev branch: cmake + catch + travis fix

aminroosta:


aminroosta:
@zauguin please review.

.ycm_extra_conf.py
```diff
@@ -0,0 +1,106 @@
...
+    '-Wall',
+    '-std=c++14',
+    '-stdlib=libc++',
+    '-x',
+    'c++',
+    '-I',
+    '.',
+    '-isystem', '/usr/local/include',
+    '-isystem', '/usr/include',
+    '-I.',
```
zauguin: The current directory is already added on line 21 and 22.

zauguin:
Is there a reason you want to add the `.ycm_extra_conf.py`? Especially the stdlib is very system specific, so I would not add a flag for this by default.

aminroosta:
Thank you for reviewing :-)
I've fixed the issues, we can merge then PRs like #126 or #125 into dev.

> Is there a reason you want to add the .ycm_extra_conf.py

I removed the flags you mentioned, feel free to update this file if you are using YCM too.
It can take a while for newcomers (who happen to be vim users) to set up.
I think there is no harm in keeping it.

PR #148, opened 2018-02-07
Separate value interface

zauguin:
No functional change intended.
This attempts to separates the conversion between C++ and SQLite types from the remaining parts of the library.
This mostly started as an experiment but I quite like it. Why?

1. Most of the type wrapping code almost directly maps to SQLite calls, so there is a lot of repeating code in the middle of our main header. I prefer to have this in a separate file to keep the main header clean.
This required some interface changes to reduce the tight coupling between the components.
2. As a sideeffect, it allows reuse of the type wrapping code without using the whole library.
3. It provides clean extension points for new types. Especially new types no langer have to add themselves as `friends` or implement user-facing interfaces like `operator <<`.
4. Also it makes `std::variant` much easier to support. Before every type needed separate handling in the `variant.h` header.

@aminroosta What do you think? Do you prefer this or our existing code?


aminroosta:
@zauguin I really like your approach. I do agree with all the points you raised.
This is awesome 👍  , reading the source code will be much easier for new contributors.

Killili:
Its hard to follow all the stuff thats happening ATM.
But i like the seperation, too. Now i just have to figure out how to support custom types again ;)

PR #149, opened 2018-02-09
Pkg-config shows wrong version of library.

ykhokhulya:


aminroosta:
@ykhokhulya 

We have switched to cmake in our [dev branch](https://github.com/SqliteModernCpp/sqlite_modern_cpp/tree/dev) which will become our next (v4) release.

However, i think we can merge this in the meantime, although automake will be completely replaced by camke in next release.

Thanks for the PR 👍

PR #153, opened 2018-02-12
Fix undefined behaviour in utf16_to_utf8

### LINKED ISSUE #151: latest build - error: Invalid UTF-16 input
The latest build is now causing the error error: Invalid UTF-16 input.
This works with prior builds ...

I am opening an existing UTF-16 database.

And this is the code that is causing the error, all inputs are std::u16string

		*awdb_ << u"SELECT value FROM settings WHERE setting = ? AND ((roaming = 0 AND machine_guid = ?) OR (roaming = 1))"
			<< setting << machine_guid_
			>> value;

zauguin:
Fixes #151.

@mtissington Please test this.

### DIFF
--- a/hdr/sqlite_modern_cpp/utility/utf16_utf8.h
+++ b/hdr/sqlite_modern_cpp/utility/utf16_utf8.h
@@ -2,6 +2,7 @@
 
 #include <locale>
 #include <string>
+#include <algorithm>
 
 #include "../errors.h"
 
@@ -10,7 +11,7 @@ namespace sqlite {
 		inline std::string utf16_to_utf8(const std::u16string &input) {
 			struct : std::codecvt<char16_t, char, std::mbstate_t> {
 			} codecvt;
-			std::mbstate_t state;
+			std::mbstate_t state{};
 			std::string result((std::max)(input.size() * 3 / 2, std::size_t(4)), '\0');
 			const char16_t *remaining_input = input.data();
 			std::size_t produced_output = 0;


mtissington:
Tested and working now - thanks!

With the changes you have implemented for converting ut16 to utf8 it would be helpful if the documentation could be updated to say a little more about this - if I open a utf16 database, it seems you process any sql as utf8 - is this safe?

zauguin:
@mtissington Yes, this is safe.
To quote the Sqlite3 source of the UTF-16 SQL Statement handler:

```c
/* This function currently works by first transforming the UTF-16
  ** encoded string to UTF-8, then invoking sqlite3_prepare(). The
  ** tricky bit is figuring out the pointer to return in *pzTail.
  */
```

So SQLite always uses UTF-8 for SQL statements, but it allows UTF-16 statements by converting them first.
This is the same behaviour this library shows, but we do this conversion by ourselves, because this gives us access to the converted string. This should never cause any change in behaviour for the user.
For this reason there is no point in using UTF-16 literals for SQL statements. This will just cause an additional conversion.
But this isn't anything related to our library, we just move this step around.

The text encoding of the database is only relevant for text strings stored in the database, so for example it decides if bound values of string types are converted.
We never touch this potential conversion, so there should never be a problem with the database encoding.

While sqlite_modern_cpp can handle UTF-16 without problems, I personally strongly recommend not to use UTF-16. For more information about the advantages of UTF-8, see http://utf8everywhere.org/.

mtissington:
Thanks for the clarity - I'm needing to use the icu library - but it makes me think, our code could be using u16string (for icu api) - you manage the input (<<) and output (>>) of parameters and results and the database could be in utf8 - or am I missing something?

zauguin:
@mtissington Right, if your database uses UTF8 encoding, you can still use u16string. The conversion will be done by SQLite.

PR #159, opened 2018-03-21
Added blob_t alternative

roussosalex:
A simple alternative to the `std::vector` approach. Makes it easier to supply a pointer and size for arbitrary binary blobs.

### DIFF
--- a/hdr/sqlite_modern_cpp/type_wrapper.h
+++ b/hdr/sqlite_modern_cpp/type_wrapper.h
@@ -36,6 +36,8 @@
 #include "errors.h"
 
 namespace sqlite {
+	using blob_t = std::pair<void const*, int>;
+
 	template<class T, int Type, class = void>
 	struct has_sqlite_type : std::false_type {};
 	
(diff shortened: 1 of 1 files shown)


aminroosta:
@roussosalex Thank you for the PR, lack of operator `>>` for raw pointers is a valid point.  
In fact we had a discussion in #122 on the interface of next version (v4).

My proposal is to change the meaning of operator `>>` for `vector<T>`.

```cpp
// single row blob
vector<int> vblob;
db << "select numbers from tbl where id = 1" >> blob_t(vblob);
// multiple row int
vector<int> vrows;
db << "select age from tbl;" >> vrows;
```

The `std::vector` is an example here, this could/should work for all standard containers.
`blob_t` can be partially specialized for different types (`blob_t<vector<T>>`, `blob_t<T*,size_t>` or `blob_t<span<T>>`).
Heres is POC for `blob_t` #125

roussosalex:
@aminroosta Thanks for the quick response.
Your proposal seems reasonable, especially because specializing `blob_t` is a much cleaner solution than overloading the `operator>>` for custom types.

PR #160, opened 2018-04-03
Catch exception objects by const reference

Peter-Levine:
Testing with latest GCC git master (8.0.1) results in errors such as:

```
tests/flags.cc: In function ‘int main()’:
tests/flags.cc:99:25: error: catching polymorphic type ‘class sqlite::sqlite_exception’ by value [-Werror=catch-value=]
  catch(sqlite_exception e)
                         ^
```

`sqlite_exception` should ideally be caught by reference (and, preferably, const reference) to avoid potential issues such as object slicing and memory exhaustion triggering further exceptions.

### DIFF
--- a/tests/blob_example.cc
+++ b/tests/blob_example.cc
@@ -52,7 +52,7 @@ int main()
 		//}
 
 	}
-	catch(sqlite_exception e)
+	catch(const sqlite_exception& e)
 	{
 		cout << "Unexpected error " << e.what() << endl;
 		exit(EXIT_FAILURE);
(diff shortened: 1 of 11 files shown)


aminroosta:
@Peter-Levine Thanks for the PR.

I think if i merge this, then we will get conflicts with dev branch.
In our dev branch we have moved the tests to use Catch2.
Also the tests in dev are not (should not be) using `try {} catch` blocks.

Could you please have a look at dev branch, we definitely need to fix it.

PR #161, opened 2018-04-03
Catch exception objects by const reference

zauguin:
A version of #160 on the `dev` branch.


zauguin:
@Peter-Levine Please review this one. It should fix all the problematic `catch` statements left in the `dev` branch.

Peter-Levine:
@zauguin The dev branch never caused failure on my end since the exceptions are caught by reference, but it's a good idea to catch by const reference.

I can confirm there is no build failure with the patch.

PR #163, opened 2018-05-19
Define __cpp_lib_uncaught_exceptions if building C++17 mode

vadz:
This notably avoids warnings about std::uncaught_exception() being
deprecated from MSVS 2017 (but note that to get it to define __cplusplus
correctly, both /std:c++17 and /Zc:__cplusplus options need to be used,
and these options are only available starting from MSVS 15.7 release).

### DIFF
--- a/hdr/sqlite_modern_cpp/utility/uncaught_exceptions.h
+++ b/hdr/sqlite_modern_cpp/utility/uncaught_exceptions.h
@@ -4,9 +4,23 @@
 #include <exception>
 #include <iostream>
 
+// Consider that std::uncaught_exceptions is available if explicitly indicated
+// by the standard library, if compiler advertises full C++17 support or, as a
+// special case, for MSVS 2015+ (which doesn't define __cplusplus correctly by
+// default as of 2017.7 version and couldn't do it at all until it).
+#ifndef MODERN_SQLITE_UNCAUGHT_EXCEPTIONS_SUPPORT
+#ifdef __cpp_lib_uncaught_exceptions
+#define MODERN_SQLITE_UNCAUGHT_EXCEPTIONS_SUPPORT
+#elif __cplusplus >= 201703L
+#define MODERN_SQLITE_UNCAUGHT_EXCEPTIONS_SUPPORT
+#elif defined(_MSC_VER) && _MSC_VER >= 1900
+#define MODERN_SQLITE_UNCAUGHT_EXCEPTIONS_SUPPORT
+#endif
+#endif
+
 namespace sqlite {
 	namespace utility {
-#ifdef __cpp_lib_uncaught_exceptions
+#ifdef MODERN_SQLITE_UNCAUGHT_EXCEPTIONS_SUPPORT
 		class UncaughtExceptionDetector {
 		public:
 			operator bool() {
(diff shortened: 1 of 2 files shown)


zauguin:
`__cpp_lib_uncaught_exceptions` is a reserved name, it should only be defined by the standard library.
This is defined in the C++ standard lex.name 3.1:

> Each identifier that contains a double underscore __ or begins with an underscore followed by an
uppercase letter is reserved to the implementation for any use.

Also if we would define this, it should have a sensible value. According to SD-6 a implementation supporting `uncaught_exceptions` as specified in N4259 should define `__cpp_lib_uncaught_exceptions` as `201411`.

That being said, we certainly need another way to detect this feature. While the SD-6 feature detection macros are a great idea, sadly libstdc++ seems to be the only standard library implementation implementing them.
Your approach of assuming the existence of C++17 library features if `__cplusplus >= 201703L` is true makes sense, but it is not exactly robust: In general most compilers set `__cplusplus` when the compiler is conformant regarding language features, not when the standard library is complete.
This is a problem especially when the compiler and the standard library are not distributed together. I recently worked on a system featuringan ancient version of libstdc++ barely supporting C++14, while a up-to-date version of clang with full C++17 support was installed.

On the other hand, systems advertising C++17 and not even supporting the most basic library features are already broken. So I would be OK with this change, iff

- A new macro is used, which is defined when `__cpp_lib_uncaught_exceptions` is defined or `__cplusplus >= 201703L` or
- Every `#ifdef __cpp_lib_uncaught_exceptions` is changed to `#if defined(__cpp_lib_uncaught_exceptions) || __cplusplus >= 201703L`

vadz:
> `__cpp_lib_uncaught_exceptions` is a reserved name, it should only be defined by the standard library.

Oops, yes, you're absolutely correct and I was too hasty. I just started using the library and wanted to have a clean build quickly, but I should have been more careful, sorry.

> While the SD-6 feature detection macros are a great idea,

Yes...

> sadly libstdc++ seems to be the only standard library implementation implementing them.

... indeed, at least I can't find any trace of them in MSVC standard library.

I'll update the PR using a new macro approach. Do you have any preferences/conventions for the name of such macro?

zauguin:
@vadz All macros we define start with `SQLITE_MODERN_CPP_`.
We also already have macros like `MODERN_SQLITE_STD_VARIANT_SUPPORT`, so it would be consistant if we would use `MODERN_SQLITE_STD_UNCAUGHT_EXCEPTIONS_SUPPORT`. But this isn't a rule, just use a name which makes the intent clear. I would probably leave out the `STD_` to shorten the name at least a little bit.

zauguin:
I'm not quite sure why, but on my system this makes the test `SEGFAULT` when build with C++17.
(ArchLinux, clang, libc++)

vadz:
Is it possible to build and run tests without using Hunter? I'd prefer to not have to deal with yet another C++ package manager if possible...

zauguin:
You can ignore my last comment, the problem was caused by a "interesting" implentation of `std::uncaught_exceptions` in `libc++` for some ABIs.

To run the tests without hunter you can set the CMake option HUNTER_ENABLED to FALSE, but then you have to manually help CMake finding sqlite3 and Catch.
This might not be necessary through: Normally there should not be a lot for you to deal with: Just use cmake for building the tests, hunter works in the background. Afterward you can just remove your build directory and all traces of hunter, the packages etc. are gone.

Issue #155, opened 2018-02-19
Multiple "warning C4065: switch statement contains 'default' but no 'case' labels" on MSVC


RunningWithScissors-:
This warning appears on multiple lines and just ruins a bit an otherwise clean build :)
I would suggest adding the following to `error_codes.h` :

```c++
#ifdef _MSC_VER
#pragma warning( push )
#pragma warning( disable: 4065 )
#endif
SQLITE_MODERN_CPP_ERROR_CODE(/**/);
/* 
.
.
.
*/
#ifdef _MSC_VER
#pragma warning( pop )
#endif
```

zauguin:
I think these pargmas belong into errors.h, because the code there actually causes the warnings.
On the other hand, we might want a more general solution here. I think there is no point in showing warnings caused by the library to users of the library.
Maybe we should wrap `sqlite_modern_cpp.h` with pragmas to suppress all warnings iff we are not building our tests.
What do you think @aminroosta @Killili

aminroosta:
> Maybe we should wrap sqlite_modern_cpp.h with pragmas to suppress all warnings iff we are not building our tests.

@zauguin Yea I think that's a good solution, but i'm afraid it may not be a common practice.

I'm also wondering how we can avoid [C4065](https://msdn.microsoft.com/en-us/library/aa748818(v=vs.60).aspx) in the first place.
Can we come up with a better macro trick to avoid this?
```cpp
switch(error_code & 0xFF) {
#define SQLITE_MODERN_CPP_ERROR_CODE(NAME,name,derived)     \
    case SQLITE_ ## NAME: switch(error_code) {          \
        derived                                           \
        default: throw name(error_code, sql); \
    }
#define SQLITE_MODERN_CPP_ERROR_CODE_EXTENDED(BASE,SUB,base,sub) \
    case SQLITE_ ## BASE ## _ ## SUB: throw base ## _ ## sub(error_code, sql);
#include "lists/error_codes.h"
#undef SQLITE_MODERN_CPP_ERROR_CODE_EXTENDED
#undef SQLITE_MODERN_CPP_ERROR_CODE
    default: throw sqlite_exception(error_code, sql);
}
```
For example instead of nested `switch` statements, we can have only two

```cpp
switch(error_code) {
   // only expand SQLITE_MODERN_CPP_ERROR_CODE_EXTENDED
   // we do not have a default
} // end of extended error_codes
switch(error_code & 0xFF) {
[...]

PR #164, opened 2018-05-19
Disable harmless MSVS warnings about switch without case

vadz:
This avoids many warnings until a better solution can be found.

See #155.


zauguin:
Thanks for the contributions.

I started another attempt to fix this with #165 and #166. If we can agree on one solution, we can avoid opening the pandoras box of filling the code with statements explicitly suppressing compiler warnings.

If this does not result in a fix in the next  few days, I will probably merge this as a temporary solution.

On another note: We switched to a model where development primary happens on the `dev` branch, so Pull Request in general should be directed towards `dev`.

zauguin:
Fixed by #165.

PR #166, opened 2018-05-20
Do not nest switch to avoid missing case ...:

### LINKED ISSUE #155: Multiple "warning C4065: switch statement contains 'default' but no 'case' labels" on MSVC

zauguin:
This is an implementation of the idea of @aminroosta to fix #155:
Instead of the current nested construction for error handling, it just uses two switch statements: One for the extended error codes, one for the normal ones.
This is an alternative to #165, so merging this would close #165.


zauguin:
Fixed by #165.

PR #167, opened 2018-06-15
Use std::string_view as argument instead of std::string on C++17 Compilers

zowpowow:
The major benefit is if someone tries to bind a `char*` parameter, we can avoid an extra copy that would happen if a new `std::string` was created from it. This pull request also includes some improvements for building/running the tests on MSVC, mostly because I suck at git and didn't know how to separate them into a separate pull request.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ OPTION(ENABLE_SQLCIPHER_TESTS "enable sqlchipher test")
 
 # Creates the file compile_commands.json in the build directory.
 SET(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-set (CMAKE_CXX_STANDARD 14)
+set (CMAKE_CXX_STANDARD 17)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
 include("cmake/HunterGate.cmake")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,7 @@ else()
 endif()
 
 catch_discover_tests(tests)
+target_compile_options(tests PUBLIC $<$<CXX_COMPILER_ID:MSVC>:/Zc:__cplusplus> )
 
 # Place the file in the source directory, permitting us to place a single configuration file for YCM there.
 # YCM is the code-completion engine for (neo)vim https://github.com/Valloric/YouCompleteMe
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -42,7 +42,7 @@ namespace sqlite {
 
 		void execute();
 
-		STR_REF sql() {
+		std::string sql() {
 #if SQLITE_VERSION_NUMBER >= 3014000
 			auto sqlite_deleter = [](void *ptr) {sqlite3_free(ptr);};
 			std::unique_ptr<char, decltype(sqlite_deleter)> str(sqlite3_expanded_sql(_stmt.get()), sqlite_deleter);
@@ -52,7 +52,7 @@ namespace sqlite {
 #endif
 		}
 
-		STR_REF original_sql() {
+		std::string original_sql() {
 			return sqlite3_sql(_stmt.get());
 		}
 
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -85,11 +85,11 @@ namespace sqlite {
 			return ++_inx;
 		}
 
-		sqlite3_stmt* _prepare(U16STR_REF sql) {
+		sqlite3_stmt* _prepare(u16str_ref sql) {
 			return _prepare(utility::utf16_to_utf8(sql));
 		}
 
-		sqlite3_stmt* _prepare(STR_REF sql) {
+		sqlite3_stmt* _prepare(str_ref sql) {
 			int hresult;
 			sqlite3_stmt* tmp = nullptr;
 			const char *remaining;
@@ -105,13 +105,13 @@ namespace sqlite {
 
 	public:
 
-		database_binder(std::shared_ptr<sqlite3> db, U16STR_REF  sql):
+		database_binder(std::shared_ptr<sqlite3> db, u16str_ref  sql):
 			_db(db),
 			_stmt(_prepare(sql), sqlite3_finalize),
 			_inx(0) {
 		}
 
-		database_binder(std::shared_ptr<sqlite3> db, STR_REF sql):
+		database_binder(std::shared_ptr<sqlite3> db, str_ref sql):
 			_db(db),
 			_stmt(_prepare(sql), sqlite3_finalize),
 			_inx(0) {
@@ -372,7 +372,7 @@ namespace sqlite {
 				*this << R"(PRAGMA encoding = "UTF-16";)";
 		}
 
-		database(const std::u16string &db_name, const sqlite_config &config = {}): database(utility::utf16_to_utf8(db_name.data()), config) {
+		database(const std::u16string &db_name, const sqlite_config &config = {}): database(utility::utf16_to_utf8(db_name), config) {
 			if (config.encoding == Encoding::ANY)
 				*this << R"(PRAGMA encoding = "UTF-16";)";
 		}
@@ -380,11 +380,11 @@ namespace sqlite {
 		database(std::shared_ptr<sqlite3> db):
 			_db(db) {}
 
-		database_binder operator<<(STR_REF sql) {
+		database_binder operator<<(str_ref sql) {
 			return database_binder(_db, sql);
 		}
 
-		database_binder operator<<(U16STR_REF sql) {
+		database_binder operator<<(u16str_ref sql) {
 			return database_binder(_db, sql);
 		}
 
@@ -419,7 +419,7 @@ namespace sqlite {
 
 			auto funcPtr = new auto(std::make_pair(std::forward<StepFunction>(step), std::forward<FinalFunction>(final)));
 			if(int result = sqlite3_create_function_v2(
-					_db.get(), name.data(), traits::arity - 1, SQLITE_UTF8, funcPtr, nullptr,
+					_db.get(), name.c_str(), traits::arity - 1, SQLITE_UTF8, funcPtr, nullptr,
 					sql_function_binder::step<ContextType, traits::arity, typename std::remove_reference<decltype(*funcPtr)>::type>,
 					sql_function_binder::final<ContextType, typename std::remove_reference<decltype(*funcPtr)>::type>,
 					[](void* ptr){
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -42,7 +42,7 @@ namespace sqlite {
 
 		void execute();
 
-		std::string_view sql() {
+		STR_REF sql() {
 #if SQLITE_VERSION_NUMBER >= 3014000
 			auto sqlite_deleter = [](void *ptr) {sqlite3_free(ptr);};
 			std::unique_ptr<char, decltype(sqlite_deleter)> str(sqlite3_expanded_sql(_stmt.get()), sqlite_deleter);
@@ -52,7 +52,7 @@ namespace sqlite {
 #endif
 		}
 
-		std::string_view original_sql() {
+		STR_REF original_sql() {
 			return sqlite3_sql(_stmt.get());
 		}
 
@@ -85,7 +85,7 @@ namespace sqlite {
 			return ++_inx;
 		}
 
-		sqlite3_stmt* _prepare(const std::u16string_view& sql) {
+		sqlite3_stmt* _prepare(const U16STR_REF& sql) {
 			//return _prepare(utility::utf16_to_utf8(sql));
 			int hresult;
 			sqlite3_stmt* tmp = nullptr;
@@ -97,7 +97,7 @@ namespace sqlite {
 			return tmp;
 		}
 
-		sqlite3_stmt* _prepare(const std::string_view& sql) {
+		sqlite3_stmt* _prepare(const STR_REF& sql) {
 			int hresult;
 			sqlite3_stmt* tmp = nullptr;
 			const char *remaining;
@@ -113,13 +113,13 @@ namespace sqlite {
 
 	public:
 
-		database_binder(std::shared_ptr<sqlite3> db, std::u16string_view const & sql):
+		database_binder(std::shared_ptr<sqlite3> db, U16STR_REF const & sql):
 			_db(db),
 			_stmt(_prepare(sql), sqlite3_finalize),
 			_inx(0) {
 		}
 
-		database_binder(std::shared_ptr<sqlite3> db, std::string_view const & sql):
+		database_binder(std::shared_ptr<sqlite3> db, STR_REF const & sql):
 			_db(db),
 			_stmt(_prepare(sql), sqlite3_finalize),
 			_inx(0) {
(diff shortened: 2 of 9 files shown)


zauguin:
Thanks for the pull request!

While this works great for users passing literals and `std::string`s, some changes are needed to also support not null-terminated `string_view`s.
Details inline.

hdr/sqlite_modern_cpp.h
```diff
@@ -85,11 +85,19 @@ namespace sqlite {
...
+		sqlite3_stmt* _prepare(const U16STR_REF& sql) {
+			//return _prepare(utility::utf16_to_utf8(sql));
+			int hresult;
+			sqlite3_stmt* tmp = nullptr;
+			const void *remaining;
+			hresult = sqlite3_prepare16_v2(_db.get(), sql.data(), -1, &tmp, &remaining);
+			if (hresult != SQLITE_OK) errors::throw_sqlite_error(hresult, utility::utf16_to_utf8(sql.data()));
+			if (!std::all_of(static_cast<const char16_t*>(remaining), sql.data() + sql.size(), [](char16_t ch) {return std::isspace(ch); }))
+				throw errors::more_statements("Multiple semicolon separated statements are unsupported", utility::utf16_to_utf8(sql.data()));
+			return tmp;
```
zauguin: Is there an advantage in not reusing the UTF-8 version?
I see some smaller problems here:

- The SQL statement in the exception is no longer the same statement given to SQLite.
In edge cases `utf16_to_utf8` might behave differently than SQLite which can make debugging hard.
- Less usage of `utf16_to_utf8` in main code paths might lead to problems in this function to go unnoticed which again makes debugging a nightmare if they only surface at error conditions.
- It is marginally slower because in case of an error the conversion happens twice.
- And of course we introduce some redundant code.

hdr/sqlite_modern_cpp.h
```diff
@@ -362,7 +370,7 @@ namespace sqlite {
 		std::shared_ptr<sqlite3> _db;
 
 	public:
-		database(const std::string &db_name, const sqlite_config &config = {}): _db(nullptr) {
+		database(const STR_REF &db_name, const sqlite_config &config = {}): _db(nullptr) {
 			sqlite3* tmp = nullptr;
 			auto ret = sqlite3_open_v2(db_name.data(), &tmp, static_cast<int>(config.flags), config.zVfs);
```
zauguin: This assumes that `db_name` is null-terminated. The `db_name` ends up as a filenamem, so not null-terminated strings are not supported here.
Maybe we should keep `std::string` here and optionally add a  `const char *` overload?

hdr/sqlite_modern_cpp.h
```diff
@@ -372,8 +380,8 @@ namespace sqlite {
 				*this << R"(PRAGMA encoding = "UTF-16";)";
 		}
 
-		database(const std::u16string &db_name, const sqlite_config &config = {}): _db(nullptr) {
-			auto db_name_utf8 = utility::utf16_to_utf8(db_name);
+		database(const U16STR_REF &db_name, const sqlite_config &config = {}): _db(nullptr) {
+			auto db_name_utf8 = utility::utf16_to_utf8(db_name.data());
```
zauguin: Same problem as above.

hdr/sqlite_modern_cpp.h
```diff
@@ -413,12 +421,12 @@ namespace sqlite {
...
 
 		template <typename Function>
-		void define(const std::string &name, Function&& func) {
+		void define(const STR_REF &name, Function&& func) {
 			typedef utility::function_traits<Function> traits;
 
 			auto funcPtr = new auto(std::forward<Function>(func));
 			if(int result = sqlite3_create_function_v2(
-					_db.get(), name.c_str(), traits::arity, SQLITE_UTF8, funcPtr,
+					_db.get(), name.data(), traits::arity, SQLITE_UTF8, funcPtr,
```
zauguin: I do not think SQlite accepts not null-terminated function names because there is no way to communicate the length.

hdr/sqlite_modern_cpp.h
```diff
@@ -427,13 +435,13 @@ namespace sqlite {
...
 		template <typename StepFunction, typename FinalFunction>
-		void define(const std::string &name, StepFunction&& step, FinalFunction&& final) {
+		void define(const STR_REF &name, StepFunction&& step, FinalFunction&& final) {
 			typedef utility::function_traits<StepFunction> traits;
 			using ContextType = typename std::remove_reference<typename traits::template argument<0>>::type;
 
 			auto funcPtr = new auto(std::make_pair(std::forward<StepFunction>(step), std::forward<FinalFunction>(final)));
 			if(int result = sqlite3_create_function_v2(
-					_db.get(), name.c_str(), traits::arity - 1, SQLITE_UTF8, funcPtr, nullptr,
+					_db.get(), name.data(), traits::arity - 1, SQLITE_UTF8, funcPtr, nullptr,
```
zauguin: See above

hdr/sqlite_modern_cpp.h
```diff
@@ -652,3 +660,5 @@ namespace sqlite {
 		}
 	}
 }
+#undef STR_REF
+#undef U16STR_REF
```
zauguin: Please add the newline at the end of the file.
Also undefining them here seems odd because they are not defined in this file.
This is not a problem for typedefs.

hdr/sqlite_modern_cpp/type_wrapper.h
```diff
@@ -4,7 +4,20 @@
...
+#define STR_REF std::string_view
+#define U16STR_REF std::u16string_view
+#else
+#define STR_REF std::string
+#define U16STR_REF std::u16string
+#endif
+#else
+#define STR_REF std::string
+#define U16STR_REF std::u16string
+#endif
```
zauguin: It is more modern and more robust to use typedefs instead of macros. It is also more consistant with `optional` etc.

hdr/sqlite_modern_cpp/type_wrapper.h
```diff
@@ -150,16 +163,15 @@ namespace sqlite {
...
 	}
 
-	// std::string
+	// STR_REF
 	template<>
 	struct has_sqlite_type<std::string, SQLITE3_TEXT, void> : std::true_type {};
-
-	inline int bind_col_in_db(sqlite3_stmt* stmt, int inx, const std::string& val) {
+	inline int bind_col_in_db(sqlite3_stmt* stmt, int inx, const STR_REF& val) {
 		return sqlite3_bind_text(stmt, inx, val.data(), -1, SQLITE_TRANSIENT);
```
zauguin: `-1` has to be replaced with the actual length.

hdr/sqlite_modern_cpp/type_wrapper.h
```diff
@@ -170,30 +182,29 @@ namespace sqlite {
 			std::string(reinterpret_cast<char const *>(sqlite3_value_text(value)), sqlite3_value_bytes(value));
 	}
 
-	inline void store_result_in_db(sqlite3_context* db, const std::string& val) {
+	inline void store_result_in_db(sqlite3_context* db, const STR_REF& val) {
 		sqlite3_result_text(db, val.data(), -1, SQLITE_TRANSIENT);
```
zauguin: `-1` has to be replaced with the actual length.

hdr/sqlite_modern_cpp/type_wrapper.h
```diff
@@ -170,30 +182,29 @@ namespace sqlite {
...
 	}
-	// std::u16string
+	// U16STR_REF
 	template<>
 	struct has_sqlite_type<std::u16string, SQLITE3_TEXT, void> : std::true_type {};
-
-	inline int bind_col_in_db(sqlite3_stmt* stmt, int inx, const std::u16string& val) {
+	inline int bind_col_in_db(sqlite3_stmt* stmt, int inx, const U16STR_REF& val) {
 		return sqlite3_bind_text16(stmt, inx, val.data(), -1, SQLITE_TRANSIENT);
```
zauguin: `-1` has to be replaced with the actual length.

hdr/sqlite_modern_cpp/type_wrapper.h
```diff
@@ -170,30 +182,29 @@ namespace sqlite {
...
 	}
-	inline std::u16string  get_val_from_db(sqlite3_value *value, result_type<std::u16string >) {
+	inline std::u16string  get_val_from_db(sqlite3_value *value, result_type<std::u16string>) {
 		return sqlite3_value_type(value) == SQLITE_NULL ? std::u16string() :
 			std::u16string(reinterpret_cast<char16_t const *>(sqlite3_value_text16(value)), sqlite3_value_bytes16(value));
 	}
 
-	inline void store_result_in_db(sqlite3_context* db, const std::u16string& val) {
+	inline void store_result_in_db(sqlite3_context* db, const U16STR_REF& val) {
 		sqlite3_result_text16(db, val.data(), -1, SQLITE_TRANSIENT);
```
zauguin: `-1` has to be replaced with the actual length.

CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ OPTION(ENABLE_SQLCIPHER_TESTS "enable sqlchipher test")
 
 # Creates the file compile_commands.json in the build directory.
 SET(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-set (CMAKE_CXX_STANDARD 14)
+set (CMAKE_CXX_STANDARD 17)
```
zauguin: We still support C++14, so it would make sense to keep this set to 14. Especially since our Travis compiler is so old that it does not support most C++17 features anyway.

hdr/sqlite_modern_cpp.h
```diff
@@ -85,11 +85,19 @@ namespace sqlite {
...
+			const void *remaining;
+			hresult = sqlite3_prepare16_v2(_db.get(), sql.data(), -1, &tmp, &remaining);
+			if (hresult != SQLITE_OK) errors::throw_sqlite_error(hresult, utility::utf16_to_utf8(sql.data()));
+			if (!std::all_of(static_cast<const char16_t*>(remaining), sql.data() + sql.size(), [](char16_t ch) {return std::isspace(ch); }))
+				throw errors::more_statements("Multiple semicolon separated statements are unsupported", utility::utf16_to_utf8(sql.data()));
+			return tmp;
 		}
 
-		sqlite3_stmt* _prepare(const std::string& sql) {
+		sqlite3_stmt* _prepare(const STR_REF& sql) {
```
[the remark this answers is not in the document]
zowpowow: Can I ask what the advantage of this would be? I know that in my use case, I'm going to have a few `string_view` objects already around and wouldn't passing by const reference be marginally faster than passing by value?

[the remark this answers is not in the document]
zauguin: In most cases they will be inlined, then it doesn't make a difference. Otherwise the advantage is that a const reference requires some object to reference, so the `string_view` has to exists in memory(typically on the stack). But when passing by value small objects like this normally are passed in registers which does not require any usage of (slow) memory.
Most of the time the string_view will not exist in memory before(because it is a temporary which did not exists or because the optimizer keeps the values in registers), so copying `2*sizeof(void*)` between registers should be much faster than writing this into memory, copying the pointer and reading from memory again.
This does mostly apply for the Itanium ABI, so not on Windows. But output from Compiler-Explorer suggests that passing by value saves a memory access on Windows too.
Additionally the interface looks cleaner :wink:

Disclaimer: I did not run a benchmark on this so I could be wrong. In this case, I would love to hear why.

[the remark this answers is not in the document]
zowpowow: You're probably right. I wrote some super crappy benchmarks and it seems passing by string_view was slightly faster: https://github.com/zowpowow/sqlite_modern_cpp/tree/str_view_by_value

zauguin:
Nice. Updating the ReadMe and adding a test would be great. There are some non-functional changes left but these are details.

It is quite sad that this doesn't really work out for function and database names. These should mostly be covered by small string optimization anyway but we could overload them for C-strings to avoid temporary strings at least when a literal is given? After merging this we might want a separate Pull Request in that direction.

hdr/sqlite_modern_cpp.h
zauguin: I think we should stick to `.c_str()` here because it makes clearer that we really need a c-style string, especially the trailing nul byte.

hdr/sqlite_modern_cpp/errors.h
```diff
@@ -9,8 +9,8 @@ namespace sqlite {
 
 	class sqlite_exception: public std::runtime_error {
 	public:
-		sqlite_exception(const char* msg, std::string sql, int code = -1): runtime_error(msg), code(code), sql(sql) {}
-		sqlite_exception(int code, std::string sql): runtime_error(sqlite3_errstr(code)), code(code), sql(sql) {}
+		sqlite_exception(const char* msg, STR_REF sql, int code = -1): runtime_error(msg), code(code), sql(std::move(sql)) {}
+		sqlite_exception(int code, STR_REF sql): runtime_error(sqlite3_errstr(code)), code(code), sql(std::move(sql)) {}
```
zauguin: The `std::move` is useless ere, moving from a const reference or a string_view is just copying.

hdr/sqlite_modern_cpp/errors.h
```diff
@@ -40,7 +40,7 @@ namespace sqlite {
 		class more_statements: public sqlite_exception { using sqlite_exception::sqlite_exception; }; // Prepared statements can only contain one statement
 		class invalid_utf16: public sqlite_exception { using sqlite_exception::sqlite_exception; };
 
-		static void throw_sqlite_error(const int& error_code, const std::string &sql = "") {
+		static void throw_sqlite_error(const int& error_code, const STR_REF &sql = "") {
```
zauguin: Here is still a const&

hdr/sqlite_modern_cpp/type_wrapper.h
```diff
@@ -31,7 +35,14 @@
...
 #endif
-
+#ifdef MODERN_SQLITE_STRINGVIEW_SUPPORT
+#include <string_view>
+typedef const std::string_view STR_REF;
+typedef const std::u16string_view U16STR_REF;
+#else
+typedef const std::string& STR_REF;
+typedef const std::u16string& U16STR_REF;
+#endif
```
zauguin: The typedef should be in out namespace to avoid introducing names in the global namespace. Also they are no longer macros, so the names should no longer be uppercase.

hdr/sqlite_modern_cpp/utility/utf16_utf8.h
```diff
@@ -8,7 +8,7 @@
 
 namespace sqlite {
 	namespace utility {
-		inline std::string utf16_to_utf8(const std::u16string &input) {
+		inline std::string utf16_to_utf8(const U16STR_REF &input) {
```
zauguin: Another const&.

zowpowow:
While we're on the topic of avoiding extra copies, is `SQLITE_TRANSIENT` really necessary over `SQLITE_STATIC`?

zauguin:
If we know we have a literal string we could use `SQLITE_STATIC` but when the user gives us a `std::string(_view)`, we can not know how long it is valid. For example using the iterator interface or prepared statements, passed temporarys will very often be destroyed before the statement is complete.
We could document the requirements and provide optional overloads for `SQLITE_STATIC` but I think the reduced safety is not worth the probably rather small performance advantage. Especially because I do not hink there is a way to detect errors here without something like `valgrind` and in most cases the database disk write will be significantly more expensive than the string copy anyway.

Of course, I would be happy to hear about ideas how this could work.

zowpowow:
Alright, Readme and tests have been updated, can we take another look at merging? Also for the `SQLITE_STATIC` thing, take a look here: https://github.com/zowpowow/sqlite_modern_cpp/commit/5d5ce076f28b5f58198b325026d63fb79d6f51d6 on how I think it could be implemented

aminroosta:
@zowpowow About SQLITE_STATIC, @5d5ce07 looks promising.
However i'm more worried about the interface, instead of having

```c++
db << "select * where name = ?" << binding_mode::static << "tom";
```

I'd recommend

```c++
db << "select * where name = ?" << sqlite::as_static("tom");
```

Please see #122 , we have a discussion on supporting blob types without `vector<T>`,
The idea was the the same to have `>> sqlite::blob(void*, size_t)` overloaded.

PR #168, opened 2018-06-19
Add binding mode option to database and database_binder

zowpowow:
Allows the user to specify to use SQLite_static instead of transient all the time. In reference to @aminroosta comment about the interface on #167, I would argue this interface is better because it's more consistent with other things in c++ that use the `<<` stream-like syntax. Iostream, for example, you can `<<` an option and it'll affect things inputted in after.  

For example:

    `std::cout  << "The number 42 in hex:     " << std::hex << 42 << '\n';`

Output: `The number 42 in hex:     2a`

At any rate, I decided to open a new pull request so we can discuss it here instead of on the old merged one.

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -27,6 +27,7 @@ namespace sqlite {
 	typedef std::shared_ptr<sqlite3> connection_type;
 
 	class row_iterator;
+
 	class database_binder {
 
 	public:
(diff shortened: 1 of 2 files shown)


zauguin:
The modifier syntax looks quite nice when used inline, but they are dangerous when the statements are passed to different code.
If a function accepts a prepared statement as argument, it can't know the current binding mode and it also can't change it without modifying the outer state.

You could add a getter to inspect the value, but bugs in functions which do not expect this are hard to find: The code looks right and it works in almost all circumstances, except in the unusual case where a user is in `do_not_copy` mode.
Also the user has to guarantee the lifetime for every passed object, so adding an explicit marker ti every object makes sense.

Issue #169, opened 2018-06-21
Not clearing all bindings


zauguin:
Often we need a statement where only some bound parameters change with every use.
Currently we clear bindings every time we reset the bindings so we have to bind all parameters again after every usage.
This gets complicated if you want to pass the prepared statement to another function without all the parameters you want to use every time.

Maybe we could make clearing the bindings optional? Or does anyone has an idea for a nice interface which allows reusing the bindings?

I thought about adding a method `.fix_parameters(n)` which indicates to no longer modify/clear the first `n` parameters. There is no `sqlite3` function for clearing a subset of all bound parameters, but we could probably emulate this.

aminroosta:
@zauguin according to documentation [sqlite3_reset](https://www.sqlite.org/c3ref/clear_bindings.html) does not rest the bindings.
The reason we are calling [sqlite3_clear_bindings](https://github.com/SqliteModernCpp/sqlite_modern_cpp/blob/master/hdr/sqlite_modern_cpp.h#L143) here. 

I wonder what happens if we don't clear the previous bindings and just bind again to an already bound parameter?

Is is ok to use [sqlite3_bind_*](https://www.sqlite.org/c3ref/bind_blob.html) on the same parameter index without clearing the existing bindings?

zauguin:
It is OK to rebind without clearing first. The only problem with it is that it can be confusing for your old bindings to still have an effect after calling reset.

PR #170, opened 2018-06-23
Support for named and indexed parameters

zauguin:
This is a UDL based attempt to support binding named parameters: For example

    db << "INSERT INTO t VALUES(:firstname, :lastname, :birthday);"
       << ":birthday"_sqlparam("2000-01-01")
       << ":firstname"_sqlparam("John")
       << ":lastname"_sqlparam("Doe");

I am not really happ with the name `_sqlparam`. Does somebody have a better idea?
Another question is how to handle mixed positional and named parameters: Currently `_sqlparam` paameters are ignored in the normal order, so for example

    db << "SELECT :abc, ?;" << ":abc"_sqlparam(1) << 2;

would bind both parameters, `1` and `2`, to `:abc` so the result would be `2, NULL`.
This is not perfect but I do not know about a better way and using both named and `?` parameters together is advised against in the SQLite manual anyway, so fixing this might not be the top priority.

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -36,17 +36,13 @@ namespace sqlite {
 		T value;
 	};
 
-	inline namespace literals {
-		inline auto operator ""_sqlparam(const char *name, std::size_t) {
-			return [name](auto &&arg) {
-				return index_binding_helper<decltype(arg), true>{name, std::forward<decltype(arg)>(arg)};
-			};
-		}
-		inline auto operator ""_sqlparam(unsigned long long index) {
-			return [index](auto &&arg) {
-				return index_binding_helper<decltype(arg)>{index, std::forward<decltype(arg)>(arg)};
-			};
-		}
+	template<class T>
+	auto named_parameter(const char *name, T &&arg) {
+		return index_binding_helper<decltype(arg), true>{name, std::forward<decltype(arg)>(arg)};
+	}
+	template<class T>
+	auto indexed_parameter(int index, T &&arg) {
+		return index_binding_helper<decltype(arg)>{index, std::forward<decltype(arg)>(arg)};
 	}
 
 	class row_iterator;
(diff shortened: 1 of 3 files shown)


zauguin:
@aminroosta The motivation are prepared statements where the parameters are not bound immediately after creating the statement.
For example I have a function which accepts a prepared statement as a parameter. During the execution the stament is executed multiple times with different values. The `?` syntax forces the caller to use the parameters inthe statement in a order implicitly defined by the callee. Also changing the implementation  such that the parameter values are known and bound in a different order becomes a API breaking change.

The caller can create a work-around by using the `?1, ?2, ...` syntax but at this point the number already acts like a name and no longer has a real meaning. So I think it is useful to allow descriptive names too.

This also gives more flexibility to the implementation of the callee.

Also the SQLite documentation says about `?` parameters: "the use of this parameter format is discouraged". The reasons do not apply for typical uses of the library, because counting the question marks is not necessary, but especially for cases where SQL statements are shared between code in different languages some support for named types might be a good idea.

This becomes even more useful in combination with #169.
On the other hand these changes but can be seen as a too big change in the fundamental design of the current API.

aminroosta:
@zauguin You make a really good point, I also agree :-)

Let's talk about the interface, I will propose to use template character packs `template<char ...c>` (or `template<char const *>`) and have an interface like this:

```c++
db << "INSERT INTO t VALUES(:firstname, :lastname, :birthday);"
   <<  named_parameter<"birthday">("2000-01-01")
   << named_parameter<"firstname">("John")
   << named_parameter<"lastname">("Doe");
```

I have not thought about the details of implementation but given we have [class template argument deduction](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction) it should be possible to implement in C++17.

What do you think?

zauguin:
@aminroosta Literal strings are not valid non-type template arguments, so AFAICT syntax like `named_parameter<":birthday">` can not be implemented. There is a proposal to change this for C++20, [P0424](https://wg21.link/p0424), but in C++17 we have to live without it.

We could create a function with two regular parameters:

    db << "INSERT INTO t VALUES(:firstname, :lastname, :birthday);"
       << named_parameter(":birthday", "2000-01-01")
       << named_parameter(":firstname", "John")
       << named_parameter(":lastname", "Doe");

zauguin:
Let's use the function then, this also allows non-literals. Then I would use `indexed_parameter` for the numbered version.

PR #172, opened 2018-11-06
Add monostate support as a nullptr

epsilon-phase:
Adding a option for the database binding operator to take `std::monostate` as a `nullptr` simplifies my code significantly. It also seems to me that it's a fairly straightforward semantic.


aminroosta:
> **std::monostate** Unit type intended for use as a well-behaved empty alternative in std::variant

@epsilon-phase I agree, i think binding `nullptr` in this case makes sense.
Few things before we merge this:

- [ ] please send this PR to `dev` branch instead
- [ ] add a line to the docs explaining this feature (in readme file where std::variant is mentioned).
- [ ] write a unit test for it (in dev branch unit tests are much easier to work with).

epsilon-phase:
Oh, Okay. Sure.

epsilon-phase:
Oh... Wow

This is very different

epsilon-phase:
Let's call this a mulligan, give me a bit for the test though :)

PR #189, opened 2019-10-09
Store error message instead of error string for exceptions

### LINKED ISSUE #187: exception handler should call sqlite3_errmsg
I've been having to keep a call to `sqlite3_errmsg` in my watch window as the exceptions never contain the actual error text. I think `throw_sqlite_error` should call `sqlite3_errmsg` and include that in its results.

zauguin:
Resolves #187.

This still makes the old sqlite3_errstr available through `e.errstr()` for applications which prefer these, but `.what()` now defaults to the error message if it is provided. Therefore this is a breaking change.


zauguin:
@BrannonKing Could you test this, please?

BrannonKing:
It runs fine for me. Thanks.

PR #191, opened 2019-10-09
Don't include experimental/optional if on Apple clang 11

shaglund:
In Apple clang 11 (Xcode 11) experimental/optional emits a warning that it has been removed, i.e building with xcode 11 fails as it seems experimental::optional is available.
This PR  fixes that by not setting MODERN_SQLITE_EXPERIMENTAL_OPTIONAL_SUPPORT if building with Apple clang 11 or above.

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -14,7 +14,7 @@
 #ifdef __has_include
 #if __cplusplus > 201402 && __has_include(<optional>)
 #define MODERN_SQLITE_STD_OPTIONAL_SUPPORT
-#elif __has_include(<experimental/optional>)
+#elif __has_include(<experimental/optional>) && __apple_build_version__ < 11000000
 #define MODERN_SQLITE_EXPERIMENTAL_OPTIONAL_SUPPORT
 #endif
 #endif


zauguin:
Thanks for your PR. I want to release the `dev` branch soon(TM), so could you move this over there? (Maybe you could also test if that branch generally works in Apple systems, especially `string_view` might have similar problems?)

Also I am very sorry, but I have to loose some words on how we got into this situation in the first place:

<details>
<summary> rant (Not about you but about these broken by design "standard libraries") </summary>

[rant on]
Why would anyone, ever provide such headers? We have lots of feature test macros in C++ such that this whole nonsense of testing the Compiler/Library version really shouldn't be necessary anymore. But that seems to be too easy. Instead, we can only use such macros after including the header (Why is that anyway???) but including the header breaks too because some people think they should provide headers which don't actually do what they promise. If Apple truly thinks that their users need a special warning and can't take the hint when their compiler says `experimental/optional not found` or similar, they can patch their compiler and add special diagnostics or patch `__has_include` to not find files which are basically empty or ... But no, everyone has to provide weird headers, making `__has_include` basically useless, the feature testing macros are hidden in the headers, making them basically useless and we are back at listing supported/unsupported Compiler version.

At this point we probably can give up any hope that this will ever change. Even if the committee would add a reliable way for feature detection, compiler/library vendors will probably just circumvent it too. 
[rant off]

</details>

But that's the world we live in, so we have to deal with it.

shaglund:
I applied the fix on the `dev` branch and it built fine in both my c++14 project and the included unit tests. I'll create a new PR for `dev`

PR #192, opened 2019-10-10
Don't include experimental/optional if on Apple clang 11

shaglund:
In Apple clang 11 (Xcode 11) experimental/optional emits a warning that it has been removed, i.e building with xcode 11 fails as it seems experimental::optional is available.
This PR fixes that by not setting MODERN_SQLITE_EXPERIMENTAL_OPTIONAL_SUPPORT if building with Apple clang 11 or above.


zauguin:
@shaglund Thank you.

PR #193, opened 2019-10-16
duplicate long type in readme

zouquan741:


aminroosta:
Screen Shot 2019-10-17 at 10 15 14 AM


It's actually a different type, `long long` is the equivalent of `int64` in C++.

zouquan741:
image
Of course I know It's a different type.
I mean there is two long type in this line, the second one should be changed to long long.
And "culumn" is a typo.

zouquan741:
I have created a new PR in dev branch with clear description and all these two problem fix, please check it.

PR #194, opened 2019-10-17
Some typo fix in README

zouquan741:
image
There is two long type in this line, the second one should be changed to long long.
And "culumn" is a typo.


aminroosta:
Thanks for the PR.

PR #204, opened 2020-02-25
upgrade hunter to newest version (to support MSVC 2019); upgrade to Catch2

marciso:
New hunter make use of `HUNTER_TLS_VERIFY` variable in cmake -- I have switched in `ON`.

For some reason after upgrading the hunter, I couldn't build because `Catch` was not found (hunter would install it but it looks like now it uses `Catch2` package instead of `Catch`...)
I figured out that `CMakeLists.txt` needs updating: `s/Catch/Catch2/`. This works for me on Windows and Linux.


aminroosta:
@marciso LGMT :+1:

PR #210, opened 2020-11-16
Added undef of __cpp_lib_uncaught_exceptions when iOS < 10.0

Szpaqn:
Added undef of __cpp_lib_uncaught_exceptions when iOS version lower than 10.0

### DIFF
--- a/hdr/sqlite_modern_cpp.h
+++ b/hdr/sqlite_modern_cpp.h
@@ -38,6 +38,12 @@
 #include <boost/optional.hpp>
 #endif
 
+#ifdef __ENVIRONMENT_IPHONE_OS_VERSION_MIN_REQUIRED__
+#if __ENVIRONMENT_IPHONE_OS_VERSION_MIN_REQUIRED__ < 100000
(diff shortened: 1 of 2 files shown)


aminroosta:
@Szpaqn Looks good, but before i merge could you explain what the background is?

Szpaqn:
Hi @aminroosta, 

I wanted to contact you, but didn't know how. 

I have an issue with Xcode version > 12 and target iOS version < 10.0.
std::uncaught_exceptions is available only for iOS 10.0 and higher, but  __cpp_lib_uncaught_exceptions is still defined even if I build for iOS 9.0 or lower.

I couldn't add a branch and create a PR in the repo, so I created a fork. 

If you could merge it to the main repository it would be great. 

Just before you do, let me change the place where the undefine is made. 
I think that the main header file (sqlite_modern_cpp.h) will be better suited.
What do you think?

Szpaqn:
@aminroosta I have changed the location for checking iOS version.
Code ready to merge from my side.

aminroosta:
Thanks @Szpaqn, Merged.

